PR #16, opened 2020-06-11
Feature/pro 33

ctucker3:
When the user tries to access a protected resource and is not logged in, or their jwt has expired, prompt them to reauthorized through orcid.

Once the user successfully authenticates, they are redirected to the page they were trying to get to. 

## Miscellaneous fixes

This also included bug fixes for the login modal:

- Login modal close button didn't work. Using `Esc` button to close caused Vue warning/error. 

Also for the system users deactivate notification messages:

- The messages still had delete language in them. I changed it to archive. 

## Dependencies

There is a partner card `feature/PRO-33` in bi-api for this card. 

## Flow
The flow differs slightly from the card acceptance criteria, the reasons will be explained below:

Flow: 
1. Unauthenticated user navigates to a protected page via an external link or manual browser navigation.
   - User will be redirected to the home page, and the login dialog will be shown with a message indicating they needed to login to see that protected resource. After they login, they are redirected to their target page. 

2. User was authenticated and on a page, but their session expires. 
    - The next api call that they try to make will cause a redirect to the home page with the login dialog shown. The login dialog will display a message that their session has expired. 
    - This is the one that differs from the card. The card says to show the login dialog on the same page. But, it was much easier to send them to home because we already had all the code for this set up. This is the behavior you see on other websites (bank websites in particular) so I think the message in the login dialog after redirect will make the process clear to users. I'm open to thoughts on this. 

## Design

bi-api was altered to look for a cookie with the name `redirect-login` on successful login. It then redirects the user to the request page instead of the default login success page. If there is no cookie with that name, the default login success url is used. This method was used instead of storing the information in local storage on the client so that we can make the login process stateless on the front end. This makes it a lot easier to manage. 

## Review

To review the different conditions, try navigating to a page that isn't the home page, 

localhost:8080/admin/program-management

this will start condition 1 above. 

Once logged in, go to a management page, open your dev tools, go to the Applications tab and delete your `phylo-token` cookie. Try to modify the data on that management page. This will start condition 2 above. 

Once logged in, open your dev tools, go to the Applications tab and delete your `phylo-token` cookie. Try to navigate to another page, this will start condition 2. 

## Shortcomings

The expired session relies on making an API call. Most of our pages will do this, but for one that doesn't an expired session user will be able to go to that page. The way to fix this is to call an endpoint (maybe an auth check url) on navigation to a new resource, which would still have the short coming of allowing them to see a resource they are already on after their session expires. Another option is to set up a repeating timed method to call an endpoint, with the downside of this being more overhead on our front end. I chose the lazy API method for now because it was the simplest and we already had infrastructure set up for it, so the pros of time savings beat out the cons. All of our sensitive information will need to be queried from the database, so the current method shouldn't be a security risk. I'm open to thoughts on this!


nickpalladino:
[PRO-33 - Reauth User](https://trello.com/c/csSlPBi4/226-pro-33-reauth-user)

src/router/index.ts
```diff
@@ -170,8 +174,19 @@ const router = new VueRouter({
 router.beforeEach((to: Route, from: Route, next: Function) => {
 
   // TODO: Check if the page is a protected resource, if not, let them through
-  // If page is protected, check if they are logged in. 
-  // TODO: Check if their token has expired.
+  // If page is protected, check if they are logged in.
+  const loginRedirectUrlCookieName = Vue.prototype.$cookieNames.loginRedirectUrl;
+
+  // Remove the redirect url from the cookie
+  Vue.$cookies.remove(loginRedirectUrlCookieName);
```
eawoods, line 181: why delete the cookie here? Didn't the redirect URL get set just two lines above?
(honest question, I've got no knowledge of Vue but it looks like this might be responsible for the no-redirect-cookie issue I'm experiencing?)

ctucker3: Fair question! The cookie is set two lines below here with, 

`Vue.$cookies.set(loginRedirectUrlCookieName, store.state.requestedPath, 60*60);`

The cookie will only exist on the home page after the non-auth redirect.

src/router/index.ts
```diff
@@ -170,8 +174,19 @@ const router = new VueRouter({
...
+
+  // Check the url for a redirect-login url. Also check if they were redirected because of expired session
+  if (store.state.requestedPath){
+    // Expires in 1 hr
+    Vue.$cookies.set(loginRedirectUrlCookieName, store.state.requestedPath, 60*60);
```
nickpalladino, line 186: So after an hour, it would just take you home rather than to the page you were on?

src/util/api.ts
```diff
@@ -30,6 +30,7 @@ export function call (config: any) {
             resolve(response);
         }).catch((error) => {
             if (error.response && error.response.status === 401) {
+                store.commit(REQUESTED_PATH, {path: window.location.href});
```
nickpalladino, line 33: Don't think this is a problem just want to make sure it's as desired. If you lose the vuex state like in a page refresh, it gets rid of login redirect and takes you back to the home page.

src/router/index.ts
```diff
@@ -195,10 +210,12 @@ router.beforeEach((to: Route, from: Route, next: Function) => {
         store.commit(ERROR_STATE, {'loginServerError':true});
       }
       // If logged in fail, send them to the home page
-      //TODO: This can go away once route protection by roles is added
```
nickpalladino, line 198: During testing if I put in a url with a program that didn't exist such as `http://localhost:8080/programs/6f895167-ff56-48e0-9675-c5fe9fe93a8a/home` it would take me to a blank page. I think a new card but we should probably handle that better.

src/router/index.ts
ctucker3: Yep! I'm not tied to an hour exactly, but wanted a short duration, but not too short, so people could remember their actions and why they are being taken where they are taken.

src/util/api.ts
ctucker3: Ah good catch, I forgot about that case. I'll check out a fix for this

src/router/index.ts
ctucker3: Yeah I noticed that behavior too. I created a new card for this:

https://trello.com/c/EQM77fiC/167-handle-state-for-page-does-not-exist

src/router/index.ts
nickpalladino: sounds ok to me

src/util/api.ts
nickpalladino: looks like this isn't reproducible and probably just a mistake on my part, sorry.

src/router/index.ts
nickpalladino: sounds good

PR #17, opened 2020-06-18
[PRO-53] show programs for users

ctucker3:
Implements changes for the system user management design card, https://cornell.app.box.com/file/674491535345. 

Is paired with a bi-api card, PRO-53, and needs a version of bi-api running that branch to work properly. 

## Misc changes

Included a way to hide columns on mobile devices in the data table. This feature was related because the admin users table takes up more space with the programs now. 

## Review 

- Make sure a program is displayed when a user is added to a program. 
- Matches the design (besides items planned to be implemented in future cards)
- We can't assign a user to >1 program in our UI yet. You can use the queries in src/test/resources/sql/UserControllerIntegrationTest.sql to assign multiple programs to a user. Do this through DBeaver.

### DIFF
--- a/src/breeding-insight/model/ProgramUser.ts
+++ b/src/breeding-insight/model/ProgramUser.ts
@@ -16,19 +16,25 @@
  * limitations under the License.
  */
 
+import {Program} from "@/breeding-insight/model/Program";
+
 export class ProgramUser {
   id?: string;
   programId?: string;
   name?: string;
   email?: string;
   roleId?: string;
+  program?: Program;
+  active?: boolean;
 
-  constructor(id?: string, name?:string, email?: string, programId?: string, roleId?: string) {
+  constructor(id?: string, name?:string, email?: string, programId?: string, roleId?: string, program?: Program, active?: boolean) {
     this.id = id;
     this.name = name;
     this.email = email;
     this.programId = programId;
     this.roleId = roleId;
+    this.program = program;
+    this.active = active;
   }
   
 }
--- a/src/breeding-insight/model/ProgramUser.ts
+++ b/src/breeding-insight/model/ProgramUser.ts
@@ -20,18 +20,16 @@ import {Program} from "@/breeding-insight/model/Program";
 
 export class ProgramUser {
   id?: string;
-  programId?: string;
   name?: string;
   email?: string;
   roleId?: string;
   program?: Program;
   active?: boolean;
 
-  constructor(id?: string, name?:string, email?: string, programId?: string, roleId?: string, program?: Program, active?: boolean) {
+  constructor(id?: string, name?:string, email?: string, roleId?: string, program?: Program, active?: boolean) {
     this.id = id;
     this.name = name;
     this.email = email;
-    this.programId = programId;
     this.roleId = roleId;
     this.program = program;
     this.active = active;
--- a/src/components/admin/AdminUsersTable.vue
+++ b/src/components/admin/AdminUsersTable.vue
@@ -24,13 +24,13 @@
     >
       <section>
         <template v-if="deactivateActive">
-          <template v-if="currentDeleteUser && currentDeleteUser.programRoles && currentDeleteUser.programRoles.length > 0">
+          <template v-if="getActiveProgramRoles().length > 0">
             <p class="has-text-dark">
               Access for this user will be removed system wide, including:
             </p>
             <ul>
               <template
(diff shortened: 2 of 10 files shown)


nickpalladino:
Question on scope, the linked box mockup shows active / inactive system user lists. I'm assuming that's in a different card somewhere?

src/components/admin/AdminUsersTable.vue
```diff
@@ -23,112 +23,183 @@
...
-          Deactivation will also deactivate the user from all programs.
-        </p>
+        <template v-if="deactivateActive">
+          <template v-if="currentDeleteUser && currentDeleteUser.programRoles && currentDeleteUser.programRoles.length > 0">
+            <p class="has-text-dark">
+              Access for this user will be removed system wide, including:
+            </p>
+            <ul>
+              <template
+                  v-for="(programRole, index) of currentDeleteUser.programRoles"
```
nickpalladino, line 33: When a user is deactivated in a program, the dialog here still shows that program in the list of programs. Seems like in that case the program should not be included in the list, only for programs they are active in.

src/components/admin/AdminUsersTable.vue
```diff
@@ -177,7 +248,8 @@ export default class AdminUsersTable extends Vue {
   }
 
   public users: User[] = [];
-  private userTableHeaders = ['Name', 'Email', 'Role'];
+  private userTableHeaders = ['Name', 'Email', 'Role', 'Programs'];
+  private hideMobileHeaders = ['Email'];
```
nickpalladino, line 252: Is hiding columns without indicating that to the user a good practice? I see the utility of being able hide certain columns for mobile but doing it without their knowledge may be confusing. Thoughts @eawoods?

src/breeding-insight/model/ProgramUser.ts
```diff
@@ -16,19 +16,25 @@
...
+  active?: boolean;
 
-  constructor(id?: string, name?:string, email?: string, programId?: string, roleId?: string) {
+  constructor(id?: string, name?:string, email?: string, programId?: string, roleId?: string, program?: Program, active?: boolean) {
     this.id = id;
     this.name = name;
     this.email = email;
     this.programId = programId;
     this.roleId = roleId;
+    this.program = program;
```
nickpalladino, line 36: Maybe at some point we can look at removing programId if we're going to have it in the program?

src/components/admin/AdminUsersTable.vue
eawoods: I agreed with Chris that removing some of the information (for now) to make the mobile version at least sort of presentable was better than a horizontally scrolling table every time.
A more thorough mobile design treatment will be forthcoming in later versions, and the information restored in some fashion, so I think it's OK for now to at least get the mobile version to be usable.

nickpalladino: Sounds good

eawoods:
> Question on scope, the linked box mockup shows active / inactive system user lists. I'm assuming that's in a different card somewhere?

Yes, inactive user lists are in PRO-57.

nickpalladino:
> > Question on scope, the linked box mockup shows active / inactive system user lists. I'm assuming that's in a different card somewhere?
> 
> Yes, inactive user lists are in PRO-57.

Cool, thanks

src/components/admin/AdminUsersTable.vue
```diff
@@ -23,112 +23,183 @@
...
+            v-on:click="deactivateActive = false"
+          >
+            Cancel
+          </button>
         </div>
       </div>              
     </WarningModal>
 
+    <div class="level">
+      <p class="level-item level-right is-text">Inactive programs are  <span class="has-background-grey-light">[ highlighted ]</span></p>
```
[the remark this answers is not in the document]
ctucker3: I had to end up going with the full `level` structure. When I did just `<p class="is-pulled-right">....</p>` it put it on the same level as the new user button. Thanks for spotting this!

src/components/admin/AdminUsersTable.vue
ctucker3: Good catch! changed and pushed

src/breeding-insight/model/ProgramUser.ts
ctucker3: I ended doing it in this pull request since the program is returned in the program user endpoints now as well.

PR #18, opened 2020-06-25
Feature/inf 81

ctucker3:
Implementation of pagination on bi-web. This is based onto PRO-53 so will include changes from that too. This should run on PRO-53 of bi-api. 

This implementation is currently mocked out, so the mocks will need to remove and functionality double checked when we have the real deal implemented. I created a card for this:

https://trello.com/c/RSYJJSKT/178-remove-mocked-pagination-from-front-end

## Implementation Details

The pagination UI exists within the `BaseTable` component, and take a `Pagination` object as a property. The `BaseTable` emits three different events, one for a page change, one for a page size change, and one for a show all toggle. It is the job of the component handling the `BaseTable` to handle these events. A `PaginationController` in the main component handles most of the data and logic for pagination to limit the amount of duplicate code in each component. When a `getAll()` request is called on a service, the response is only assigned to the data table if the returned response matches the most recent request. I did it this way so that in the case that a request takes longer than it does for the user to click 'Next', we are not lagging behind the users actions. 

Only the `getAll()` requests implement pagination, the other requests will have no need for them.

You will see a large chunk of css in the main.css. I went with the buefy pagination implementation, so needed to override some css in order to get it looking the right way. 

For the mocks, the calls are still made as normal, but the data is parsed on the front end to get the pagination effect. if you don't want to enter a ton of records to see the pagination in action, you can use this query to generate a bunch of locations. If you want more than 100 records, change the max number in the series generator

```
insert into program(name, species_id, created_by, updated_by)
select 'Test Program Locations', species.id, bi_user.id, bi_user.id from
species 
join bi_user on bi_user."name" = 'system'
limit 1;

insert into place(name, program_id, coordinates, created_by, updated_by)
select concat('test_location', i::text), program.id, 'null', program.created_by, program.updated_by from generate_series(1, 100) s(i)
join program on name = 'Test Program Locations'
```

## A note and question:

Our current pagination implements a 0 indexed `currentPage` property. We initially did this to match brapi. Staying with this on our backend means more logic on the front end. I wonder if we want to change our backend implementation to be 1 starting index to make it more translatable to the UI. I don't see much advantage in doing 0 index for a pagination result parameter. Anyone have thoughts on this? 

## Testing

- Play around with the pagination controls and make sure it works as you expect. 
- Open your dev tools and look at the `network` tab. Interact with the pagination controls and check that the request that is sent reflects the action you made in the pagination controls. This is necessary because our backend doesn't do pagination yet, so all results are returned regardless of the request we send. 
- Pagination controls match the design specification
- Pagination controls are not shown when there is no data in the table

### DIFF
--- a/src/assets/scss/main.scss
+++ b/src/assets/scss/main.scss
@@ -135,7 +135,8 @@ nav.pagination {
     display: flex;
     flex-grow: 1;
     flex-shrink: 0;
-    justify-content: space-evenly;
+    justify-content: flex-start;
+    margin-left: 2em;
 
     & div.page-size-select {
       order: 1;
--- a/src/assets/scss/main.scss
+++ b/src/assets/scss/main.scss
@@ -101,7 +101,7 @@ main {
   pre,
   table {
     &:not(last-child) {
-    margin-bottom: 1em;
+      margin-bottom: 1em;
     }
   }
 }
@@ -110,9 +110,44 @@ main ul {
   list-style: disc outside;
   margin-left: 2em;
   ul {
-      list-style-type: circle;
-      margin-top: .5em;
+    list-style-type: circle;
+    margin-top: .5em;
+  }
+}
+
+
+
+nav.pagination {
+  align-items: normal;
+  justify-content: flex-start;
+  & ul.pagination-list {
+    list-style-type: none;
+    margin-left: 0;
+    flex-grow: 0;
+    flex-shrink: 0;
+    & li a.is-current {
+      background-color: $info;
+      border: none;
+    }
+  }
+  & div.pagination-extras {
+    order: 4;
+    display: flex;
+    flex-grow: 1;
+    flex-shrink: 0;
+    justify-content: space-evenly;
+
+    & div.page-size-select {
+      order: 1;
+      border: none;
+      & span {
+        margin-left: 4px;
       }
+    }
+    & a.show-all-button {
+      order: 2;
+    }
+  }
 }
 
 main ol {
--- a/src/components/program/ProgramUsersTable.vue
+++ b/src/components/program/ProgramUsersTable.vue
@@ -107,9 +107,13 @@
       v-bind:records.sync="users"
       v-bind:row-validations="userValidations"
       v-bind:editable="true"
+      v-bind:pagination="usersPagination"
       v-on:submit="updateUser($event)"
       v-on:remove="displayWarning($event)"
       v-on:show-error-notification="$emit('show-error-notification', $event)"
+      v-on:paginate="paginationController.updatePage($event)"
+      v-on:paginate-toggle-all="paginationController.toggleShowAll()"
+      v-on:paginate-page-size="paginationController.updatePageSize($event)"
     >
       <template v-slot:columns="data">
         <TableRowColumn name="name">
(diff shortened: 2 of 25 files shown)


src/assets/scss/main.scss
```diff
@@ -110,9 +110,44 @@ main ul {
...
-      list-style-type: circle;
-      margin-top: .5em;
+    list-style-type: circle;
+    margin-top: .5em;
+  }
+}
+
+
+
+nav.pagination {
```
nickpalladino, line 120: Regarding the layout, things are a bit more spread out than in the mockups, not sure if it really matters but a difference. I didn't see anything in the mockup for button highlight color, not sure if it's supposed to match the table row highlight color in the mockup? @eawoods

src/components/program/ProgramUsersTable.vue
```diff
@@ -215,10 +225,19 @@ export default class ProgramUsersTable extends Vue {
     this.getUsers();
   }
 
+  @Watch('paginationController', { deep: true})
   getUsers() {
 
-    ProgramUserService.getAll(this.activeProgram!.id!).then((programUsers: ProgramUser[]) => {
```
nickpalladino, line 220: When testing in with the programUsers table, if I add a new user (first user to exist) then deactivate them, the DELETE is successful and the GET is successful with no program user but the data table is not updated unless I reload the page.

ctucker3: Good catch! It was an error with mocking. Fixed and pushed

src/assets/scss/main.scss
ctucker3: The reasoning behind that was to show the user the current state of the pagination. It should be an easy fix if you want different behavior @eawoods? Thanks for bringing this up Nick!

eawoods: Layout: yeah, maybe not so spread out and more matching the design. Best way would be setting that `justify-content` on line 138 to 'flex-start` and then adding `.pagination-extras .pagination-link { margin: 2em }` to get the closer-but-not-too-close spacing.

For the current-page color: the `$info` color is exactly as intended! If I didn't illustrate it, you read my mind.

ctucker3:
> * When user is on any page that is not page 1, creating a new item highlights the top row of the current page, but the new item is on page 1.
>   When I did this with users, I had ten users and added an eleventh -- that worked fine, but the locations table had the highlighting on the wrong page (the current page). Or....also, adding an eleventh user (new page) did NOT take that user to the top of the list, but the top of the new page. I hope that's making sense. It looks like the logic is being expressed as:
>   -- Lots of items, multiple pages with current page not full: add new item to top of list on first page, highlight top row of current page (new item does not show with highlighting)
>   -- Lots of items, user is on last page which is full: add new item to top of NEW page and highlight it (so it looks like the correct behavior except that the new item doesn't go to the first page as expected).
>   -- Further testing: I take that back, the highlighting and new-item creation order are now doing very strange things. Sometimes not appearing at the top of either list, sometimes not getting highlighted....there's something funky going on and I don't see the whole pattern.

Liz and I discussed this, and it this is due to the backend not sorting the results based on created date. I created a mocked sorting in bi-web so it behaves as if bi-api had sorted the data. NOTE: The admin users table still doesn't sort because that endpoint doesn't return the createdAt date. I also added a navigation back to page 1 when a new item was created so a user could see that the row was created. We will have to give this some more thought when data table sorting functionality is added. 

There is also a bug in the data table when highlighting new rows from this as well. When you are on a page other than page 1 and create a new item, it will navigate you back to page 1, but the new row will not be highlighted. The fix will require a total rework of the data table new item highlighting logic, so I think it best to put off for another card:

https://trello.com/c/1FXRz1Me/179-bug-data-table-highlighting-doesnt-work-with-pagination

src/assets/scss/main.scss
ctucker3: changed and pushed

PR #20, opened 2020-06-30
[INF-138] Remove workaround for data not existing in response

ctucker3:
Removes workarounds for the `data` object not being included in the response if no results were in the data response. Goes with bi-api INF-138.


ctucker3:
[INF-138 - API result for GET /programs not returning an empty data array](https://trello.com/c/UPReRYjh/239-inf-138-api-result-for-get-programs-not-returning-an-empty-data-array)

PR #21, opened 2020-07-20
ONT-21 - Import via Excel workbook or CSV

nickpalladino:
## Changes
- Added Traits sidebar menu items with placeholder view pages
- Added file selector component
- Added progress bar component
- Added trait upload service
  - PUT / GET
- Added imported traits datatable
- Added Import traits page

## Running
- New dependencies were added for xstate & bulma 0.9 so you'll need to do `npm install` prior to running.
- You'll need the latest develop branch of bi-api

## Design
- Import traits page most state management done with with [XState](https://xstate.js.org/docs/)
  - [State chart diagram](https://xstate.js.org/viz/?gist=c55b667010f06a607f0d83b457028246)
  - XState visualizer only supports javascript so just converted on the [typescript playground](https://www.typescriptlang.org/play)
  - More verbose but formalizing helps me manage more complex state
    - Left modal management out
- Tried to break out file importing components so they can be reused as-is in other file import pages
- Table loading state for when there is a lot of data (mostly an issue due to pagination mocking currently for large files)
- Bulma has some new [spacing helpers](https://bulma.io/documentation/helpers/spacing-helpers) which are very useful

## Testing
- I created a large trait file with 50,0001 traits for testing the progress bar and table loading
  - File available [here](https://cornell.box.com/s/jakpca4yiqpm43h1scqg2tst9smaf2h9)
  - Other test files available in the [bi-api repo](https://github.com/Breeding-Insight/bi-api/tree/develop/src/test/resources/files)

## Acceptance Criteria
- See [ONT-21 trello card](https://trello.com/c/gJAtAvm7) for the acceptance criteria
  - New cards were created due to the narrowed scope of ONT-21


nickpalladino:
[ONT-21 - Import via Excel workbook or CSV](https://trello.com/c/gJAtAvm7/227-ont-21-import-via-excel-workbook-or-csv)

nickpalladino:
> Overall the import function performs well and as expected, using Firefox/MacOS.
> 
> _Known issue_: on MacOS Mojave 10.14.6 using Chrome v.84.0.4147.89, any `.xls` files return an 'Unsupported mime type' error. This is because Chrome/Mac returns type `application/octet-stream` rather than the expected `application/vnd.ms-excel` type. However, `.csv` and `.xlsx` files can be imported (and throw proper errors for missing columns, etc) as expected.
> 
> Nick and I talked, and the plan is to write a new card for this browser-specific and type-specific bug (because the fix is backend rather than front-end).

Added card for this: https://trello.com/c/4KAHzhuG

src/breeding-insight/model/Scale.ts
```diff
@@ -0,0 +1,30 @@
...
+
+export class Scale {
+  scaleName?: string;
+  dataType?: string;
+  categories?: Array<Category>;
+
+  constructor(scaleName?:string, dataType?:string, categories?:Array<Category>) {
+    this.scaleName = scaleName;
+    this.dataType = dataType;
+    this.categories = categories;
```
ctucker3, line 28: How about decimal places, valid value min, valid value max? Or maybe these aren't in our design yet?

src/breeding-insight/model/Trait.ts
```diff
@@ -0,0 +1,40 @@
...
+              programObservationLevel?: ProgramObservationLevel,
+              method?: Method,
+              scale?: Scale
+              ) {
+    this.traitName = traitName;
+    this.programObservationLevel = programObservationLevel;
+    this.method = method;
+    this.scale = scale;
+  }
+}
```
ctucker3, line 40: Missing description, traitClass, attribute, defaultValue, mainAbbreviation, synonyms, abbreviations.

src/breeding-insight/service/TraitUploadService.ts
```diff
@@ -0,0 +1,86 @@
...
+import {PaginationQuery} from "@/breeding-insight/model/PaginationQuery";
+import {PaginationController} from "@/breeding-insight/model/view_models/PaginationController";
+
+export class TraitUploadService {
+
+  static uploadFile(programId: string, file: File): Promise<ProgramUpload> {
+
+    return new Promise<ProgramUpload>((resolve, reject) => {
+
+      if (file !== null) {
```
ctucker3, line 31: Maybe should check if programId is also not null?

src/components/file-import/FileSelector.vue
```diff
@@ -0,0 +1,79 @@
...
+    }
+
+  }
+</script>
+
+<!-- keeping this scoped since it's only used here -->
+<style scoped>
+.file-select > input[type="file"] {
+  display: none;
+}
```
ctucker3, line 78: Do we want to start adopting this style for all of our components with custom styling? It might keep our main css file from accumulating a bunch of them.

src/views/TraitsImport.vue
```diff
@@ -0,0 +1,287 @@
...
+          >
+            Cancel
+          </button>
+        </div>
+      </div>              
+    </warning-modal>
+
+    <template v-if="state === ImportState.CHOOSE_FILE || state === ImportState.FILE_CHOSEN">
+      <h1 class="title">Import Traits</h1>
+      <trait-import-template-message-box class="mb-5"/>
```
ctucker3, line 50: Does this need to be a separate component since it is specific to traits? Perhaps this could be generalized to work with all types of imports?

src/components/trait/TraitImportTemplateMessageBox.vue
```diff
@@ -0,0 +1,57 @@
...
+              <div class="has-text-dark">
+                <strong>Before You Import...</strong>
+                <br/>Prepare trait information for import using the provided template.
+              </div>
+            </div>
+          </div>
+          <div class="level-right">
+            <div class="level-item">
+              <div class="has-text-dark has-text-centered is-size-7">
+                <!-- temporary link until the backend card is done, need github access to our private repo -->
```
ctucker3, line 34: Is there a card for this?

timparsons: Could this template (for now) be in either bi-api or bi-web?

ctucker3:
You will need to assign your user to the program for this to work, to assign your user, run the following query in the DB, 

```
insert into program_user_role (user_id, program_id, role_id, created_by, updated_by)
select bi_user.id, program.id, role.id, bi_user.id, bi_user.id from bi_user
join program on program.name = 'New Program 2'
join role on role."domain" = 'breeder'
where bi_user."name" = 'BI-DEV Admin'
```

src/breeding-insight/model/Scale.ts
nickpalladino: Ya, these are minimal just for what I needed and slightly extra but not the entire models. Figured we'd just fill them out as we needed to add stuff.

src/breeding-insight/model/Trait.ts
nickpalladino: Same as above, just put in stuff I needed for now.

src/components/file-import/FileSelector.vue
nickpalladino: I think it makes sense for on-off things that aren't going to be reused anywhere as it keeps our main css clear of clutter.

src/breeding-insight/service/TraitUploadService.ts
nickpalladino: Ya, I'll add it.

src/views/TraitsImport.vue
nickpalladino: I mainly broke it out to keep the TraitsImport page free of clutter and easier to read. I'm not sure if the formatting is going to be the same for different file import types @eawoods? If it is, we could change the name and add props for customizing. If it's all going to be different I wouldn't bother with a slot and would just make one for each type. It'd probably be better to wait until we get the template file loading stuff in there to figure out the proper interface.

src/components/trait/TraitImportTemplateMessageBox.vue
nickpalladino: Ya, card is https://trello.com/c/1FvZFLch. Probably be served out of backend once we get to that card. For now I can put the file as a public link in box maybe? I could serve it from the front end but think it would be easier to have a public link.

nickpalladino:
> Nice work! it meets all the acceptance criteria when I run it. One minor quip: I missed in my initial once-over the asterisks showing required fields in the downloaded template, so maybe highlight those columns to visually call attention to them.
> 
> We talked already about eventually using magic numbers to validate the excel file format. That could be done eventually as part of the file picker or just on the backend or both. For reference, the card I created for that is INF-186.

Thanks, I added a card to highlight required field in the file template: https://trello.com/c/dEqYwUpo. I linked your card and my card for the file validation since they overlap.

src/breeding-insight/service/TraitUploadService.ts
nickpalladino: Pushed changes.

src/views/TraitsImport.vue
nickpalladino: From discussion in meeting, we're going to hold off on doing anything with this for now.

PR #23, opened 2020-07-28
[INF-75] add ability to add existing user to program

ctucker3:
Modifies the front end to be able to send add an existing user to the program. 

### Design

I modified the front end instead of the backend so that we could keep our desired behavior for the endpoints in the front end. When an acting user adds a user to a program, the front end checks if a system user with that email exists, and gets the id from that user if so. The front end then uses its existing methods to POST the user. When the response returns, we check if the id of the returned user matches the id of one of the existing users, and shows a notification indicating an existing user was added to the program. 

### How to test

- Look in the system users table and copy one of the ids there. 
- Create a few programs
- Go into each of the programs and add a user with the copied email above
- When you save, make sure the notification clearly indicates that an existing user was added to the program
- After adding the user to all of the programs, go back to the system users page and confirm that all of the user's programs are listed in that table.

### DIFF
--- a/src/components/program/ProgramUsersTable.vue
+++ b/src/components/program/ProgramUsersTable.vue
@@ -183,6 +183,8 @@ import EmptyTableMessage from "@/components/tables/EmtpyTableMessage.vue";
   import {PaginationController} from "@/breeding-insight/model/view_models/PaginationController";
   import {PaginationQuery} from "@/breeding-insight/model/PaginationQuery";
   import {Pagination} from "@/breeding-insight/model/BiResponse";
+  import { User } from '@/breeding-insight/model/User';
+  import {UserService} from "@/breeding-insight/service/UserService";
 
 @Component({
   mixins: [validationMixin],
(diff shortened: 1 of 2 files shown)


src/components/program/ProgramUsersTable.vue
```diff
@@ -223,6 +226,7 @@ export default class ProgramUsersTable extends Vue {
   mounted() {
     this.getRoles();
     this.getUsers();
+    this.getSystemUsers();
```
nickpalladino, line 229: Do you think the system users should be read on create as well in case something changed between loading the component and the request? Probably unlikely but could potentially be out of date if things changed in the mean time? I guess there could be similar things with roles and the data shown in the drop downs but that would be a little more difficult to deal with I think.

nickpalladino:
[INF-75 - Existing User Add to Program](https://trello.com/c/XmXdQpco/255-inf-75-existing-user-add-to-program)

src/components/program/ProgramUsersTable.vue
ctucker3: I was thinking about doing it that way, but this was simpler and a temporary fix with limited users using it so went with this route. I'm definitely down with switching to the way you described if we get complaints of users being created when they already exist though. Are you cool with waiting, or do you think we should do that now?

nickpalladino: Ya, I think it's fine for now. Just something to pay attention to in other places as well as we work on stuff. Since we're planning on updating this anyways with the search stuff we can just wait.

PR #24, opened 2020-07-31
[PRO-78] manual entry of orcids

ctucker3:
Adds a manual entry of the orcid to the admin users and program users endpoints. 

## Design
- Most extra orcid logic is added to the services and daos for easy removal. There is some stuff in the view pages that will have to be removed as well. 

## Behavior
- In Program Users, if the orcid (or email) matches an existing (and active) user, that existing user will be added to the program. 

## Known Limitations
- In Admin Users, if you add a valid user with a duplicate user id, there will be an error message that the user was created, by the orcid was a duplicate. The user will be created, but the id will not be assigned. You can assign this later. 
- Same goes for updating a user in admin users. 
- In Program Users, if an orcid overlaps the orcid of an inactive user, it will not add the inactive user to the program, it will show a duplicate error. A new user will be created without an orcid in this case.

### DIFF
--- a/src/components/program/ProgramUsersTable.vue
+++ b/src/components/program/ProgramUsersTable.vue
@@ -211,7 +211,7 @@ export default class ProgramUsersTable extends Vue {
   public users: ProgramUser[] = [];
   public systemUsers: User[] = [];
   private usersPagination?: Pagination = new Pagination();
-  userTableHeaders: string[] = ['Name', 'Email', 'Orcid', 'Role'];
+  userTableHeaders: string[] = ['Name', 'Email', 'Role'];
 
   private deactivateActive: boolean = false;
   private newUserActive: boolean = false;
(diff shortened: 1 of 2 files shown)


src/components/program/ProgramUsersTable.vue
```diff
@@ -203,12 +211,13 @@ export default class ProgramUsersTable extends Vue {
   public users: ProgramUser[] = [];
   public systemUsers: User[] = [];
   private usersPagination?: Pagination = new Pagination();
-  userTableHeaders: string[] = ['Name', 'Email', 'Role'];
+  userTableHeaders: string[] = ['Name', 'Email', 'Orcid', 'Role'];
```
nickpalladino, line 214: Added header here for orcid but no column in table data.

ctucker3: Changed and pushed

PR #26, opened 2020-08-06
ONT-37 - View individual trait & ONT-43 - View uploaded trait before finalizing import

nickpalladino:
## Changes
- Have two data table types now, expandable row and side panel
- Updated existing admin & program tables to use expandable row table
- Updated trait import and trait list tables to use side panel table
- See https://trello.com/c/hEz1c0Ri for detailed breakdown of subtasks
- Fixed some other small existing issues
  - got rid of unwanted title tool tip on pages
  - fixed alignment of user login text with logout button
  - removed some space between page header and content title to be more inline with design reference

## Running
- A new dependency was added for vue breakpoint so you'll need to run `npm install`

## Design
- See https://trello.com/c/hEz1c0Ri for high level design
- Table column model based on buefy data table (not latest version)
  - Looked into vue provide/inject as an alternative way to get table reference but held off for now due to issues
- Table breakpoint column collapse is done dynamically with vue breakpoint

## Limitations
- Uses position: sticky css which limits browser support (particularly IE)
  - May eventually switch to a site wide contextual side panel and address this (https://trello.com/c/r0Z2P5k0)
- Sidepanel set to 100vh to allow for panel content scrolling when sticky and table bigger than sidepanel
  - Should probably make the close button not scroll at some point
  - May eventually switch to a site wide panel as stated above
- In mobile, page header covers part of sticky side panel
  - should fix it if we get some time
- Only supporting existing scale types (Code, Date, Duration, Nominal, Numerical, Ordinal, Text) for trait details panel
  - design reference has extra types such as boolean, integer, units
- Parser only allows integers for min/max values, not dates which it seems should be allowed?
  - Should we create a card for this?

## Testing
- Created test file for testing display of traits detail side panel for different method/scale types and scrolling
- [test_trait_details.xlsx](https://github.com/Breeding-Insight/bi-web/files/5035019/test_trait_details.xlsx)

## Acceptance Criteria
- See https://trello.com/c/hEz1c0Ri & https://trello.com/c/sj0tsBx2


nickpalladino:
[ONT-43 - View uploaded trait before finalizing import](https://trello.com/c/sj0tsBx2/244-ont-43-view-uploaded-trait-before-finalizing-import)

nickpalladino:
[ONT-42 - Trait Import Failed](https://trello.com/c/xa1YZ15E/245-ont-42-trait-import-failed)

nickpalladino:
[ONT-37 - View individual trait](https://trello.com/c/hEz1c0Ri/247-ont-37-view-individual-trait)

ctucker3:
> Parser only allows integers for min/max values, not dates which it seems should be allowed?
Should we create a card for this?

BrAPI has the min/max scale values as integers, so allowing dates in there too might cause some issues. Maybe adding our own custom fields for mix/max date would play better with BrAPI?

src/components/tables/BaseTable.vue
```diff
@@ -17,189 +17,91 @@
...
-            slot="next"
-            slot-scope="props"
-        >
-          <b-pagination-button
-              :page="props.page"
-              tag="a"
+      <thead v-if="updatedColumns.length">
+        <tr>
+          <!-- Header space for left row icon if desired -->
+          <th v-if="showRowIcon" width="40px"/>
```
ctucker3, line 29: What is the left row icon?

src/components/tables/BaseTable.vue
```diff
@@ -228,52 +130,42 @@
...
-              }
-            }
-          }
-        }
-      }
-
-      return {}
-    }
+    addColumn(column: TableColumn) {
+      const repeated = this.updatedColumns.some(
```
ctucker3, line 134: It took me a bit of digging to figure out how/where this was being used. Might be useful for future reference to have a comment saying that it is used by the TableColumn to add itself to the table.

src/components/tables/BaseTable.vue
```diff
@@ -228,52 +130,42 @@
...
+      if (!repeated) {
+        this.updatedColumns.push(column);
+      }
     }
 
-    getValidations(index: number) {
-      return this.$v.tableRows.$each[index].editData;
+    removeColumn(column: TableColumn) {
+      if (!this.tableRows.length) return;
+      if (this.updatedColumns.length !== 1) return;
```
ctucker3, line 144: I'm a little confused by this method. Why is this condition necessary? I see that below you are seeing if the column is in the table's columns and removing if so. But wouldn't this condition prevent removal if there is more than one column in the table? 

Also, same comment as for the `addColumn` might be helpful for future reference to have a comment saying it is used by the TableColumn component.

src/components/tables/BaseTable.vue
```diff
@@ -17,189 +17,91 @@
...
+    showExpandControls!: boolean;
+    @Prop()
+    showRowIcon!: boolean;
+    
+    private initialUpdate: boolean = false;
+    private previouslyVisible: Array<TableColumn> = [];
     private tableRows: Array<TableRow<any>> = new Array<TableRow<any>>();
+    private updatedColumns: Array<TableColumn> = [...this.columns];
+    private isMobile = false;
+    private isTable = true;
```
ctucker3, line 86: This variable doesn't seem to be used anywhere.

src/components/tables/BaseTable.vue
```diff
@@ -17,189 +17,91 @@
...
-    pagination!: Pagination;
-
-    initialUpdate: boolean = false;
-
+    showExpandControls!: boolean;
+    @Prop()
+    showRowIcon!: boolean;
+    
+    private initialUpdate: boolean = false;
+    private previouslyVisible: Array<TableColumn> = [];
```
ctucker3, line 82: This variable doesn't seem to be used anywhere.

src/components/tables/ExpandableRowTable.vue
```diff
@@ -0,0 +1,108 @@
...
+      <slot name="emptyMessage" />
+    </template>
+
+  </div>
+</template>
+
+<script lang="ts">
+
+  import {Component, Prop, Vue, Watch, Mixins} from 'vue-property-decorator'
+  import {TableRow} from "@/breeding-insight/model/view_models/TableRow"
```
ctucker3, line 83: Extra import

src/components/tables/ExpandableRowTable.vue
```diff
@@ -0,0 +1,108 @@
...
+  import BaseTable from '@/components/tables/BaseTable.vue'
+  import {Pagination} from "@/breeding-insight/model/BiResponse";
+  import PaginationControls from '@/components/tables/PaginationControls.vue'
+  import ExpandableTableRow from "@/components/tables/ExpandableTableRow.vue"
+  import EditDataRowForm from '@/components/forms/EditDataRowForm.vue'
+  import {Validations} from "vuelidate-property-decorators";
```
ctucker3, line 89: Extra import

src/components/tables/ExpandableRowTable.vue
```diff
@@ -0,0 +1,108 @@
...
+  - See the License for the specific language governing permissions and
+  - limitations under the License.
+  -->
+
+<template>
+  <div>
+    <!-- need to have base-table loaded for watching records 
+         for new highlighting so use v-show 
+    -->
+    <base-table
```
ctucker3, line 23: Should this be BaseTable to be consisting with the naming of our other components?

src/components/tables/ExpandableRowTable.vue
```diff
@@ -0,0 +1,108 @@
...
+                  name="edit"
+                />
+              </EditDataRowForm>
+            </td>
+          </tr>
+        </template>
+      </template>
+
+      <slot v-for="(_, name) in $slots" :name="name" :slot="name" />
+      <template v-for="(_, name) in $scopedSlots" :slot="name" slot-scope="slotData"><slot :name="name" v-bind="slotData" /></template>
```
ctucker3, line 69: What do these do? Is it sending all slots from BaseTable up to the parent component? Is the row slot included among the slots shuffled up?

src/components/tables/SidePanelTable.vue
```diff
@@ -0,0 +1,199 @@
...
+    @Prop()
+    editable!: boolean;
+    @Prop()
+    pagination!: Pagination;
+
+    private panelOpen = false;
+
+    private collapseColumns: boolean = false;
+    private columns: Array<any> = [];
+    private CollapseColumnsState = CollapseColumnsState;
```
ctucker3, line 109: The variables above this look unused

src/components/tables/SidePanelTable.vue
```diff
@@ -0,0 +1,199 @@
...
+    private BreakpointEvent = BreakpointEvent;
+    private PanelEvent = PanelEvent;
+    private CollapseColumnAction = CollapseColumnAction;
```
ctucker3, line 112: Do you need to assign these to variables if you've already declared them in enums above?

src/components/tables/TableColumn.vue
```diff
@@ -0,0 +1,81 @@
...
+    private visible!: boolean;
+
+    @Prop({ default: false })
+    private collapsed!: boolean;
+
+    @Prop()
+    private label!: string;
+
+    @Prop([Number, String])
+    private width!: number | string | undefined;
```
ctucker3, line 45: Is there a benefit of this being allowed to be a string? Looks like it can only integer like when it is used in the BaseTable

```
v-bind:style="{
                width: column.width === undefined ? null :
                (isNaN(column.width) ? column.width : column.width + 'px')
              }"
```

src/components/tables/TraitDetailPanel.vue
```diff
@@ -0,0 +1,169 @@
+<!--
```
ctucker3, line 1: This component doesn't seem to belong to the `tables/` group of components. Should this be in the 'traits/' directory?

src/components/trait/TraitListsTable.vue
```diff
@@ -131,10 +133,9 @@ export default class TraitTable extends Vue {
   private newTrait = new Trait();
   private traitName: string = "Program Trait";
   private deleteTrait?: Trait;
```
ctucker3, line 135: `deleteTrait`, `newTrait`, and `newTraitActive` seem to no longer be used.

src/mixins/ValidationMixin.ts
```diff
@@ -0,0 +1,71 @@
+/*
```
ctucker3, line 1: I don't remember if this worked before, but the invalid field messages no longer appear in the expandable edit section forms. They still work for the new data form. Attaching screen shots of what I see.

screencapture-localhost-8080-admin-program-management-2020-08-07-13_53_52
screencapture-localhost-8080-admin-program-management-2020-08-07-13_53_38

src/components/tables/BaseTable.vue
nickpalladino: I left that in there for when we do the data table highlighting for the error or warning icons.

src/components/tables/SidePanelTable.vue
```diff
@@ -0,0 +1,199 @@
...
+              />
+            </SidePanelTableRow>
+          </template>
+
+          <slot v-for="(_, name) in $slots" :name="name" :slot="name" />
+          <template v-for="(_, name) in $scopedSlots" :slot="name" slot-scope="slotData"><slot :name="name" v-bind="slotData" /></template>
+        </base-table>
+      </div>
+      <div v-bind:class="{'column is-narrow is-gapless pl-0': !panelOpen, 'column is-one-third-desktop is-half-tablet is-half-mobile is-gapless pl-0': panelOpen}" >
+        <side-panel v-show="records.length > 0" class="side-panel-scroll" v-if="panelOpen" v-on:close-panel="closePanel" v-bind:background-color-class="'has-background-info-light'">
```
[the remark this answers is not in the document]
nickpalladino: @eawoods can correct me if I'm wrong but I believe the design reference color needs to be updated to match the info-light color generated by buefy.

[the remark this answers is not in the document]
eawoods: It is correct in the code and is probably too dark in the design reference. The Bulma "light" colors are calculated, and sometimes I guess wrong on how light that function goes.

src/components/tables/SidePanelTable.vue
```diff
@@ -0,0 +1,199 @@
...
+          <slot v-bind:data="selectedRow.data" name="side-panel"/>
+        </side-panel>
+      </div>
+    </div>
+    <pagination-controls v-show="records.length > 0" v-bind="$props" v-on="$listeners"/>
```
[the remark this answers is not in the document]
nickpalladino: Could you point me to that file so I can try it out?

src/components/tables/BaseTable.vue
nickpalladino: Sounds good, I'll add coments

src/components/tables/SidePanelTable.vue
nickpalladino: Ya, it wasn't working unless I created those in the class. If you know of a way around that, let me know. I had a similar issue with some classes with static methods similar to what we do for the services so I'm not sure why it wasn't working.

src/components/tables/TraitDetailPanel.vue
nickpalladino: Ya, that sounds good, I'll move it.

src/components/tables/BaseTable.vue
nickpalladino: I'll add a comment for that. It's used by the TableColumn for finding the table.

src/components/tables/BaseTable.vue
nickpalladino: Thanks, I'll get rid of that.

src/components/tables/ExpandableRowTable.vue
nickpalladino: Does IntelliJ show you these? VSCode isn't which makes them hard to catch.

src/components/trait/TraitListsTable.vue
nickpalladino: Thanks, I'll remove those

src/mixins/ValidationMixin.ts
nickpalladino: I'll check this out, so it still gives you the notification but not the field errors?

src/components/tables/ExpandableRowTable.vue
ctucker3: Yep! Gotta upgrade to the pro version!

src/mixins/ValidationMixin.ts
ctucker3: Yep, exactly.

src/components/VBreakpoint.js
```diff
@@ -0,0 +1,35 @@
...
+  },
+  debounceTime: 100,
+  experimental: false
+}
+ 
+const { VShowAt, VHideAt, VBreakpoint } = extend(config)
+ 
+export default VBreakpoint
+ 
+export { VShowAt, VHideAt, VBreakpoint }
```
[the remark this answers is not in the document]
nickpalladino: I think we could have different configured versions, have to look into it. This configuration is to match the bulma breakpoints, not really specific to the data table. Maybe could be named to reflect that but would we need another set of viewport breakpoints? I think if we move to the layout sidebars we could use ResizeObserver for changes in content size?

nickpalladino:
> > Parser only allows integers for min/max values, not dates which it seems should be allowed?
> > Should we create a card for this?
> 
> BrAPI has the min/max scale values as integers, so allowing dates in there too might cause some issues. Maybe adding our own custom fields for mix/max date would play better with BrAPI?

Ya, the min/max date is coming from crop ontology which BrAPI based stuff off. You can see in the BrAPI spec under the min value it says for numerical and date scales. It may be an oversight in BrAPI if the intention is to actually support that?

src/components/tables/ExpandableRowTable.vue
nickpalladino: Thanks I'll remove that.

src/components/tables/ExpandableRowTable.vue
nickpalladino: Ya, for some reason I got the idea that kebab case was better to use in the templates but looking into it again, I should just stick with pascal case so I'll change this.

src/components/tables/ExpandableRowTable.vue
nickpalladino: These are left over from earlier work and should be removed because I'm no longer passing slots through to the BaseTable.

src/components/tables/SidePanelTable.vue
nickpalladino: Thanks I'll remove collapseColumns and columns.

src/components/tables/TableColumn.vue
nickpalladino: I guess just flexibility, if you pass a string it'll just pass it through but if you give a number it will add 'px' to it. If anything I'd probably keep string over number.

src/breeding-insight/utils/StringFormatters.ts
```diff
@@ -0,0 +1,27 @@
...
+export class StringFormatters {
+  
+  static toStartCase(type: string) : string | undefined {
+    if (type) {
+      return type.charAt(0).toUpperCase() + type.substr(1).toLowerCase();
+    }
+    return undefined;
+  }
+
+}
```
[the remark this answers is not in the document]
nickpalladino: I'll test this but I think in those cases it would return an empty string. Maybe it should return undefined instead.

src/components/tables/SidePanelTable.vue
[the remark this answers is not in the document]
nickpalladino: The file in the breedbase channel didn't import so I made some changes to it to get it to import (github's not letting me upload it for some reason). The pagination seems to be working fine for me. If you have a different file or configuration that causes the issue you were seeing, let me know.

timparsons, line 48: I tried uploading a trait file, and changing the pagination to show 20 at a time, selected an item, navigated to the next page, then back, and was expecting to see the original item still be the highlighted row, but did not.  Not sure if this has already been decided on, but wanted to point it out in case it hasn't.

src/components/tables/TraitDetailPanel.vue
```diff
@@ -0,0 +1,169 @@
...
+    }
+
+    get abbreviationsSynonymsString() {
+      var abbSyn = "";
+      if (this.trait && this.trait.abbreviations && this.trait.abbreviations.length > 0) {
+        abbSyn = this.trait.abbreviations[0];
+      }
+      if (this.trait && this.trait.synonyms && this.trait.synonyms.length > 0) {
+        abbSyn = (abbSyn === "") ? this.trait.synonyms[0] : abbSyn + ", " + this.trait.synonyms[0];
+        abbSyn = this.trait.synonyms.length > 0 ? abbSyn + ", ...": "";
```
timparsons, line 127: This last line will put an extra `, ...` even if there is only 1 synonym for the trait

src/components/trait/TraitsImportTable.vue
```diff
@@ -17,31 +17,49 @@
...
+          {{ TraitStringFormatters.getScaleTypeString(data.scale) }}
+        </TableColumn>
       </template>
+
+      <!-- 
+        Side panel data slot specification
+        data: T
+      -->
+      <template v-slot:side-panel="data">
+        <trait-detail-panel v-bind:data="data"/>
```
timparsons, line 57: Should this tag also be updated to be pascal case instead of kebab?

src/components/trait/TraitListsTable.vue
```diff
@@ -35,45 +35,45 @@
...
-            />
-          </div>
-        </div>
+
+      <!-- 
+        Side panel data slot specification
+        data: T
+      -->
+      <template v-slot:side-panel="data">
+        <trait-detail-panel v-bind:data="data"/>
```
timparsons, line 74: Should this tag also be updated to be pascal case instead of kebab?

src/components/VBreakpoint.js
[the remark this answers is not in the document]
ctucker3: I think if it matches the bulma breakpoints it is fine as it is. I imagine the design will be tailored to the different device viewports.

src/components/tables/BaseTable.vue
ctucker3: Nice, that makes sense.

src/components/tables/SidePanelTable.vue
ctucker3: Ok, I figured it was probably something that didn't work if you didn't add them. Maybe the Vue component needs them in its scope to be able to interact with them? Either way, looks good to me, no changes needed.

ctucker3:
> Ya, the min/max date is coming from crop ontology which BrAPI based stuff off. You can see in the BrAPI spec under the min value it says for numerical and date scales. It may be an oversight in BrAPI if the intention is to actually support that?

Ah ok, I just looked at the schema, not the description. Maybe something we want to check with Pete on before we commit to support date ranges? I'm thinking if BrAPI intentionally does not support date ranges in that field, it will add a lot more complexity/logic and potential breakages for that field if we allow dates in there.

src/components/tables/SidePanelTable.vue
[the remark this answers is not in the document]
ctucker3: Hrrm trying it again, I can't reproduce the pagination not working. So, nevermind, this is good!

src/breeding-insight/utils/StringFormatters.ts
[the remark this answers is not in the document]
ctucker3: I'm cool with either way.

src/components/tables/SidePanelTable.vue
```diff
@@ -0,0 +1,199 @@
+<template>
+  <div class="side-panel-table">
+    <div class="columns is-mobile">
+      <div class="column pr-0">
+        <base-table
```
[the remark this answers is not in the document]
nickpalladino: Ya, I'll update that

src/components/tables/TraitDetailPanel.vue
nickpalladino: Good catch. @eawoods I took some liberties with this as it wasn't in the design. Are you ok with doing this after maybe more than 1 or 2 synonyms so it doesn't end up going to the next line and looking weird at some point?

src/components/trait/TraitsImportTable.vue
nickpalladino: Yep, I'll update it

src/components/trait/TraitListsTable.vue
nickpalladino: Ya. I'll update that

src/components/tables/SidePanelTable.vue
eawoods: https://trello.com/c/1FXRz1Me/205-inf-179-bug-data-table-highlighting-doesnt-work-with-pagination

src/components/tables/TraitDetailPanel.vue
eawoods: @nickpalladino yep, that sounds good to me.

src/components/tables/SidePanelTable.vue
nickpalladino: As @eawoods pointed out there is a card for new data highlighting. We should either add to the scope of that card or create a new card for keeping table rows open (in the case of expandable row table) or table rows selected and open (in the case of sidepanel) if those are actual requirements.

nickpalladino:
> > Ya, the min/max date is coming from crop ontology which BrAPI based stuff off. You can see in the BrAPI spec under the min value it says for numerical and date scales. It may be an oversight in BrAPI if the intention is to actually support that?
> 
> Ah ok, I just looked at the schema, not the description. Maybe something we want to check with Pete on before we commit to support date ranges? I'm thinking if BrAPI intentionally does not support date ranges in that field, it will add a lot more complexity/logic and potential breakages for that field if we allow dates in there.

Ya, I'll ask Pete about it.

nickpalladino:
> > > Ya, the min/max date is coming from crop ontology which BrAPI based stuff off. You can see in the BrAPI spec under the min value it says for numerical and date scales. It may be an oversight in BrAPI if the intention is to actually support that?
> > 
> > 
> > Ah ok, I just looked at the schema, not the description. Maybe something we want to check with Pete on before we commit to support date ranges? I'm thinking if BrAPI intentionally does not support date ranges in that field, it will add a lot more complexity/logic and potential breakages for that field if we allow dates in there.
> 
> Ya, I'll ask Pete about it.

Created card https://trello.com/c/3QrdgElK and issue on BrAPI github. Pete's going to change the field type to support dates so we'll wait for that and have this card for when it's ready.

src/breeding-insight/utils/StringFormatters.ts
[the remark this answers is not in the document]
nickpalladino: For empty it returns undefined and for length 1 it just returns the first letter capitalized. If we think it should change in the future we can. For right now it's only used for formatting scale type which has a lot of validation on it such that there shouldn't be strings like this.

src/components/tables/BaseTable.vue
nickpalladino: Pushed changes.

src/components/tables/BaseTable.vue
nickpalladino: Pushed changes.

src/components/tables/BaseTable.vue
nickpalladino: Pushed changes.

src/components/tables/ExpandableRowTable.vue
nickpalladino: Pushed changes.

src/components/tables/ExpandableRowTable.vue
nickpalladino: Pushed changes.

src/components/tables/ExpandableRowTable.vue
nickpalladino: Removed from ExpandableRowTable and SidePanelTable. Pushed changes.

src/components/tables/SidePanelTable.vue
timparsons: What I put in the comment was just what I had expected it to do given that the detail panel stays open when I switch to a different page of the datatable, I don't recall that being explicitly defined as a requirement.  @eawoods I defer to you on that.

src/components/tables/ExpandableRowTable.vue
nickpalladino: Pushed changes.

src/components/tables/SidePanelTable.vue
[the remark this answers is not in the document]
nickpalladino: Pushed changes.

src/components/trait/TraitsImportTable.vue
nickpalladino: Pushed changes.

src/components/trait/TraitListsTable.vue
nickpalladino: Pushed changes.

src/breeding-insight/utils/StringFormatters.ts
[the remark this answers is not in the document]
ctucker3: Sounds good to me!

src/components/tables/SidePanelTable.vue
nickpalladino: Ah, I see what you're saying now. Maybe it should just close the panel on any pagination event since it loses the highlighting (related to the new row highlighting I think). That would be similar to how the expandable row table closes and unhighlights on any pagination event. @eawoods @timparsons

src/components/tables/SidePanelTable.vue
nickpalladino: Pushed changes.

src/components/tables/SidePanelTable.vue
timparsons: I'm fine with that

src/components/tables/TraitDetailPanel.vue
nickpalladino: Pushed changes.

src/components/trait/TraitListsTable.vue
nickpalladino: Pushed changes.

src/components/tables/TraitDetailPanel.vue
nickpalladino: Pushed changes.

src/components/tables/SidePanelTable.vue
eawoods: Yep, me too - closing the panel when switching pages is consistent with other behaviors as @nickpalladino mentioned.

nickpalladino: Just to clarify "pagination events" you could resize and still have the item on the page but it would still close at least for now as pagination is resetting the tablerow state. Hopefully that's alright for now.

nickpalladino: Pushed changes.

nickpalladino:
> Looks fantastic!!!! The thoughtful application of scrolling in the side panel for long scales is especially nice.
> 
> One tiny detail, if it isn't too much to request: in the side panel is it possible to map the scale methods to a phrase? Examples:
> "Ordinal" => "Ordinal Scale"
> "Nominal" => "Nominal Scale"
> "Numerical" => "Numerical Units"
> (for the line "Plant Observation in Ordinal" to become "Plant Observation in Ordinal Scale")
> ...some of them wouldn't get a word, like "Date" or "Text" probably. I had slightly different language in the design reference due to having slightly different list of scales (still couldn't find where the heck I got that from).
> 
> If it's too much trouble, then this is a small point and not worth holding up the card for more than a short time. It would read more cleanly with the extra word to make a coherent noun, but is not a make-or-break item.

Pushed changes.

src/components/tables/BaseTable.vue
nickpalladino: Pushed changes.

PR #32, opened 2020-08-19
[ISSUE-6] fix wave errors

ctucker3:
Fixed the issues from the trello card. There were also issues in the new items form and edit items form that were fixed. I checked the trait import to see if there were any errors there, it was all good. 

## Dependencies

Can be run with develop of bi-api.

### DIFF
--- a/src/components/tables/BaseTable.vue
+++ b/src/components/tables/BaseTable.vue
@@ -38,7 +38,7 @@
           </th>
           <!-- Add a header column to match spacing for row controls if specified -->
           <template v-if="showExpandControls">
-            <th />
+            <td />
           </template>
         </tr>
       </thead>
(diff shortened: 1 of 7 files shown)


nickpalladino:
[ISSUE-6 - WAVE accessibility scan shows errors in general header](https://trello.com/c/3Iv54L9T/272-issue-6-wave-accessibility-scan-shows-errors-in-general-header)

src/components/tables/BaseTable.vue
```diff
@@ -38,7 +38,7 @@
           </th>
           <!-- Add a header column to match spacing for row controls if specified -->
           <template v-if="showExpandControls">
-            <th />
+            <td />
```
nickpalladino, line 41: Was there an issue associated with this?

eawoods: Accessible table guidelines say that table headers should not be empty, in the reference I'm reading. https://webaim.org/techniques/tables/data 
Since `th` and `td` elements are being mixed in this row, the recommendation is to add `scope="col"` to each of the `th` entries (good practice to do so anyway).

PR #35, opened 2020-08-21
INF-199 - Create unit tests for data tables in bi-web

ctucker3:
Implements tests for the data tables. Also did some cleanup, removing mocha from the unit test dependencies since we are using jest, and cleaning up other unused packages in the package.json. 

## Design

Uses specific tables (ex. Program locations) instead of testing the table components directly because the configuration is easier to set up. It is a pain to try and set up all the slots and data and events for all of the tests. 

## Tests Implemented

See the trello card, https://trello.com/c/zox4BocQ/271-inf-199-create-unit-tests-for-data-tables-in-bi-web, for the tests that were implemented. 

## Review

You shouldn't need to pull this down, just a code review will do. I added the unit tests to github actions, so if they are passing then running the tests in your local environment shouldn't be necessary.


nickpalladino:
[INF-199 - Create unit tests for data tables in bi-web](https://trello.com/c/zox4BocQ/271-inf-199-create-unit-tests-for-data-tables-in-bi-web)

tests/unit/components/tables/ExpandableRowTable.spec.ts
```diff
@@ -0,0 +1,284 @@
...
+    expect(editBtn.exists()).toBeTruthy();
+    let editForm = wrapper.findComponent(EditDataRowForm);
+    expect(editForm.exists()).toBeFalsy();
+    await editBtn.trigger('click');
+
+    editForm = wrapper.findComponent(EditDataRowForm);
+    expect(editForm.exists()).toBeTruthy();
+  });
+
+  it('Displays input error when form validation occurs', async () => {
```
nickpalladino, line 68: Do we want to add a test that would catch issues like that state one that we had where it would work the first time and then not after that?

tests/unit/components/tables/ExpandableRowTable.spec.ts
```diff
@@ -0,0 +1,284 @@
...
+    await openEditForm(wrapper);
+
+    const nextPageBtn = wrapper.find('a[aria-label="Next page"');
+    await nextPageBtn.trigger('click');
+    await wrapper.vm.$nextTick();
+
+    const row = wrapper.findComponent(ExpandableTableRow);
+    const firstRowName = row.find('td[name="name"]');
+    expect(firstRowName.text()).toEqual('Test100');
+  });
```
nickpalladino, line 247: Should this also check that the expandable row is closed?

tests/unit/components/tables/SidePanelTable.spec.ts
```diff
@@ -0,0 +1,231 @@
...
+  it('Next page button works when details open', async () => {
+    await openEditForm(wrapper);
+
+    const nextPageBtn = wrapper.find('a[aria-label="Next page"');
+    await nextPageBtn.trigger('click');
+    await wrapper.vm.$nextTick();
+
+    const row = wrapper.findComponent(SidePanelTableRow);
+    const firstRowName = row.find('td[name="name"]');
+    expect(firstRowName.text()).toEqual('Trait100');
```
nickpalladino, line 193: Should this also check that the sidepanel is closed?

tests/unit/components/tables/SidePanelTable.spec.ts
```diff
@@ -0,0 +1,231 @@
...
+  });
+});
+
+async function openEditForm(wrapper: Wrapper<any>) {
+  const row = wrapper.findComponent(SidePanelTableRow);
+  const editBtn = row.find('a[data-testid="showDetails"]');
+  await editBtn.trigger('click');
+  let editForm = wrapper.findComponent(SidePanel);
+  expect(editForm.exists()).toBeTruthy();
+}
```
nickpalladino, line 231: Not sure if we want to add some tests for the breakpoints column collapse events?

tests/unit/components/notifications.spec.ts
```diff
@@ -49,7 +49,7 @@ describe('notifications display', () => {
     expect(wrapper.html()).not.toContain(successMsg);
 
     // Emit event
-    wrapper.find(UserManagement).vm.$emit('show-success-notification', successMsg);     
+    wrapper.find(AdminUserManagement).vm.$emit('show-success-notification', successMsg);
```
nickpalladino, line 52: getting warnings about find being deprecated for columns and should use findComponent

tests/unit/components/tables/ExpandableRowTable.spec.ts
ctucker3: I would vote to not add that test now, but add it later if that bug comes up again. Since it is fixed, it's going to be hard to write a test to reproduce that state. I think it would be easier and more useful to write those specific tests so that it fails when the bug is present, and then passes when you fix the bug.

tests/unit/components/notifications.spec.ts
ctucker3: Ah cool, thanks for finding that! I saw those warning too but couldn't find the culprit. I'll change that.

tests/unit/components/tables/SidePanelTable.spec.ts
ctucker3: Do you mean going the opposite way as well then? Desktop -> tablet -> mobile?

ctucker3: Check the `collapse` and `uncollapse-columns` events emitted by SidePanelTable.

tests/unit/components/tables/ExpandableRowTable.spec.ts
nickpalladino: Works for me

tests/unit/components/tables/SidePanelTable.spec.ts
ctucker3: Changed and pushed. I tried to capture switching between viewport states and panel states, and change in panel states while staying in the same state. See a list of test descriptions below:
```
✓ Collapses columns when panel is opened and change to mobile (45ms)
    ✓ Uncollapses columns when on mobile and panel closed (41ms)
    ✓ Collapses columns when on mobile and panel opened (27ms)
    ✓ Columns stay collapsed when panel is opened and switch to tablet (1ms)
    ✓ Uncollapses colums when on tablet and panel closed (41ms)
    ✓ Collapses when on tablet and panel is open (23ms)
    ✓ Uncollapses when panel opened and switch to desktop (20ms)
    ✓ Stays uncollapsed when panel is closed (27ms)
```

tests/unit/components/tables/SidePanelTable.spec.ts
ctucker3: Changed and pushed

tests/unit/components/tables/ExpandableRowTable.spec.ts
ctucker3: Changed and pushed

tests/unit/components/notifications.spec.ts
ctucker3: Changed and pushed

PR #41, opened 2020-09-09
[ISSUE-8],[BI-540] Fix orcid overwrite when adding new program user for existing system user

ctucker3:
Based off bug/PRO-80, so review and approve that one first. 

This bug fix fixes a bug where you can update the orcid when creating a program user for a system user that already exists. Previously, if there was an email match, the orcid would be updated for the existing user. Bug fix makes it so the update orcid action is not made when there is an email match. 

Also adds a not documented bug where if the email and orcid you enter reference two different users, it would arbitrarily choose one. I made it so it throws an error if this condition is met. 

screencapture-localhost-8080-programs-6d7a6ab8-4b25-4cc1-b572-4531efd0533b-program-management-users-2020-09-10-09_27_07
screencapture-localhost-8080-programs-6d7a6ab8-4b25-4cc1-b572-4531efd0533b-program-management-users-2020-09-10-09_27_37
screencapture-localhost-8080-programs-6d7a6ab8-4b25-4cc1-b572-4531efd0533b-program-management-users-2020-09-10-09_28_18

### DIFF
--- a/src/breeding-insight/service/ProgramUserService.ts
+++ b/src/breeding-insight/service/ProgramUserService.ts
@@ -28,8 +28,8 @@ export class ProgramUserService {
 
   static errorEmailInUse: String = "Error creating user, a user with this email already exists";
   static errorCreatingUser: String = "Error while creating user";
-  static errorAssignedUserToProgramErrorOrcid = "Successfully assigned user to program, error assigning orcid to user.";
-  static errorOrcidRequired = "Orcid required when creating a new user to add to the program";
+  static errorAssignedUserToProgramErrorOrcid = "Successfully assigned user to program, error assigning ORCID iD to user.";
+  static errorOrcidRequired = "ORCID iD required when creating a new user to add to the program";
 
   static create(programUser: ProgramUser, orcid: string | undefined): Promise<ProgramUser> {
 
--- a/src/breeding-insight/service/UserService.ts
+++ b/src/breeding-insight/service/UserService.ts
@@ -34,10 +34,10 @@ export class UserService {
   static errorDeleteUserNotFound: string = 'Unable to find user to deactivate';
   static errorDeleteUserNotAllowed: string = 'You are not allowed to deactivate this user.';
   static errorPermissionsEditUser: string = "You don't have permissions to edit the roles of this user.";
-  static errorUpdatingOrcidOnPost: string = "User created, but could not assign orcid";
-  static errorUpdatingOrcidOnPut: string = "User updated, but could not update orcid";
-  static errorUpdatingOrcidOnPostDuplicate: string = "User created, but could not assign orcid. Orcid Id already in use.";
-  static errorUpdatingOrcidOnPutDuplicate: string = "User updated, but could not update orcid. Orcid Id already in use.";
+  static errorUpdatingOrcidOnPost: string = "User created, but could not assign ORCID iD";
+  static errorUpdatingOrcidOnPut: string = "User updated, but could not update ORCID iD";
+  static errorUpdatingOrcidOnPostDuplicate: string = "User created, but could not assign ORCID iD. ORCID iD Id already in use.";
+  static errorUpdatingOrcidOnPutDuplicate: string = "User updated, but could not update ORCID iD. ORCID iD Id already in use.";
 
   static getUserInfo(): Promise<User> {
 
--- a/src/components/admin/AdminUsersTable.vue
+++ b/src/components/admin/AdminUsersTable.vue
@@ -127,7 +127,7 @@
             <BasicInputField
                 v-model="newUser.orcid"
                 v-bind:validations="validations.orcid"
-                v-bind:field-name="'Orcid'"
+                v-bind:field-name="'ORCID iD'"
             />
           </div>
           <div class="column is-one-fourth">
@@ -162,7 +162,7 @@
           {{ data.email }}
         </TableColumn>
         <!--TODO: Remove when registration flow is complete -->
-        <TableColumn name="orcid" v-bind:label="'Orcid'" v-bind:visible="!isMobile">
+        <TableColumn name="orcid" v-bind:label="'ORCID iD'" v-bind:visible="!isMobile">
           {{ data.orcid }}
         </TableColumn>
         <TableColumn name="roles" v-bind:label="'Role'">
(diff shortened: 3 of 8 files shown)


eawoods:
Tested this out. As expected, when entering a random ORCID iD for an existing user (following the "steps to reproduce"), that new and erroneous ORCID does not propagate to the system. However, there is no indication in the UI that the new ORCID was ignored. Do we care about that? I'm inclined to say no, because I'd really like this entire feature to go away sooner rather than later. It's a liability and has been sprouting a number of bugs.

Nice find on the "double match" case -- is the error message that's thrown in this case "Error creating user, a user with this email already exists"?  (If so then I am confirming the intended result.)

Again I'm not anxious to quibble on wording or any fine points, as it would be far better to have done as soon as possible.

ctucker3:
> Tested this out. As expected, when entering a random ORCID iD for an existing user (following the "steps to reproduce"), that new and erroneous ORCID does not propagate to the system. However, there is no indication in the UI that the new ORCID was ignored. Do we care about that? I'm inclined to say no, because I'd really like this entire feature to go away sooner rather than later. It's a liability and has been sprouting a number of bugs.

I would also vote to leave the message as it is. We will be setting up the accounts for our sandbox users, so the lack of indication that the ORCID was ignored is low impact, in my opinion. 

> Nice find on the "double match" case -- is the error message that's thrown in this case "Error creating user, a user with this email already exists"? (If so then I am confirming the intended result.)

No, that is the wrong error message. It should say, "Error Email and Orcid match two different users". I've attached screen shots of the different states above. Can you send the steps you took where the wrong error message came up?

eawoods:
My bad -- I had gotten the email address wrong when I was trying to match name/email and then add a random ORCID iD. When I fixed the email address so that it was the one matching the name of the user I was trying to add, I got the "Email and Orcid match two different users" message as outlined above. So, never mind that report from above that I was getting the wrong message - it was user error.

P.S. I *will* quibble over the proper spelling of "ORCID" and that the actual term for the number is "ORCID iD", should be used both on data collection labels and in error messages.

timparsons (changes requested):
I noticed that there is quite a bit of business logic that's on the front end, for example with the `checkExistingUserByEmailOrOrcid` function.  With this function in particular, I noticed that it's using a browser-side cached list of system users to evaluate if a user already exists with the ORCID iD/email (instead of fetching fresh from the db when attempting to add a new user).  I did the following test:

1. In one browser window, open the program users screen 
1. on another window open the system users screen
1. create a new system user
1. Go back to the program users screen and attempt to create a new user with the same email but different ORCID iD

I was trying to see if this would result in the user's ORCID iD being overwritten, of which is was not.  However, the error message I got was `Error creating user, a user with this email already exists `, meaning the backend logic prevented me from creating the user, but I should have been allowed to add the user to the program because they have the same email.

This may be a case we are ok with for now given that a) ORCID iD input will go away once the full user registration flow is implemented, and b) we will have a small amount of users, so concurrent creation of additional users is likely to be small.

Last comment on business logic in the UI, I didn't extensively go through the UI codebase, so what I observed may only be applicable to the ORCID iD input, however, it would be worth assessing how much business logic is in the UI now, and moving it to the API (this is outside the scope of this fix).

ctucker3:
> I noticed that there is quite a bit of business logic that's on the front end, for example with the `checkExistingUserByEmailOrOrcid` function. With this function in particular, I noticed that it's using a browser-side cached list of system users to evaluate if a user already exists with the ORCID iD/email (instead of fetching fresh from the db when attempting to add a new user). 
> This may be a case we are ok with for now given that a) ORCID iD input will go away once the full user registration flow is implemented, and b) we will have a small amount of users, so concurrent creation of additional users is likely to be small.

Yep, I agree that this isn't a good final solution and that we want to keep business logic in the backend for our permanent non-workaround features. I explained the reason for adding this to the front end instead of modifying the backend in this pull request:

https://github.com/Breeding-Insight/bi-web/pull/23

I was thinking the same thing with your points a and b, in addition my intention was to make the update orcid parts easily removable, to keep our existing backend endpoints unmodified since the email match and orcid update are temporary workarounds, and to move in a general direction toward what our final implementation will be. Searching and adding an existing user through the search menu will be able to use the current setup of sending an id back, without the `checkExistingUserByEmailOrOrcid` method. Updating the orcid in the registration flow will be able to use the PUT orcid endpoint that was added to the backend.

I did the following test:
> 
> 1. In one browser window, open the program users screen
> 2. on another window open the system users screen
> 3. create a new system user
> 4. Go back to the program users screen and attempt to create a new user with the same email but different ORCID iD
> 
> I was trying to see if this would result in the user's ORCID iD being overwritten, of which is was not. However, the error message I got was `Error creating user, a user with this email already exists `, meaning the backend logic prevented me from creating the user, but I should have been allowed to add the user to the program because they have the same email.
> 

Your test case makes sense, I believe that is the same scenario Nick tested out in the above PR as well. A simple solution would be to do a fresh pull of the system users immediately before the create call, but with the limited users we decided we could wait on adding that until it becomes an issue. That isn't bad to add though if you want me to add it to this PR so we don't run into this issue?

> Last comment on business logic in the UI, I didn't extensively go through the UI codebase, so what I observed may only be applicable to the ORCID iD input, however, it would be worth assessing how much business logic is in the UI now, and moving it to the API (this is outside the scope of this fix).

I believe every permanent feature in our UI has the business logic in the backend. The orcid input, and adding existing user to a program are the exceptions because they are temporary workarounds. Doesn't hurt to look through other stuff though.

ctucker3:
> P.S. I _will_ quibble over the proper spelling of "ORCID" and that the actual term for the number is "ORCID iD", should be used both on data collection labels and in error messages.

I've updated all references to ORCID to be ORCID iD.

timparsons:
> Your test case makes sense, I believe that is the same scenario Nick tested out in the above PR as well. A simple solution would be to do a fresh pull of the system users immediately before the create call, but with the limited users we decided we could wait on adding that until it becomes an issue. That isn't bad to add though if you want me to add it to this PR so we don't run into this issue?

Nah, I don't think that's worth it either at this point in time.  Thanks for the detailed response, it was very helpful to me.  I'll approve this PR as it is.

timparsons (approved):

eawoods (approved):
Approved - thank you !

PR #42, opened 2020-09-10
[BI-72] - Error adding new user with already existing orcid, form stays open

ctucker3:
Fixes the issue of the new data form being left open when a user is successfully saved but their orcid is not updated properly. 

Additions:
- Added a duplicate orcid check before making POST check.
- Closed new data form if user POST is successful, but orcid PUT is not. 
- Added validation for orcid
- Changed addUser method to use async/await for clearer error handling and request flow. 

Note: 

Updating the users uses a similar request flow to what adding users used to be before updating to use async/await. I did not change update users because I did not want this card to get too out of scope. 

## Limitations

The pre-check for a duplicate orcid does not take into account deactivated users. Our API is not capable of returning inactive users yet. The backend will still catch duplicates for inactive users though. See attached error state. 

Bad ORCiD iD
screencapture-localhost-8080-admin-user-management-2020-09-10-14_53_03
Duplicate orcid precheck
screencapture-localhost-8080-admin-user-management-2020-09-10-14_52_48
Duplicate email
screencapture-localhost-8080-admin-user-management-2020-09-10-14_53_38
Duplicate orcid with inactive user
screencapture-192-168-10-72-8080-admin-user-management-2020-09-18-11_45_43

### DIFF
--- a/src/App.vue
+++ b/src/App.vue
@@ -21,6 +21,7 @@
       <SuccessNotification ref="successNotification" class="is-marginless"></SuccessNotification>
       <ErrorNotification ref="errorNotification" class="is-marginless"></ErrorNotification>
       <InfoNotification ref="infoNotification" class="is-marginless"></InfoNotification>
+      <WarningNotification ref="warningNotification" class="is-marginless"></WarningNotification>
     </div>
     
     <component v-bind:is="layout" v-bind:username="username" @logout="logOut">
@@ -28,6 +29,7 @@
             @show-success-notification="showSuccessNotification"
             @show-info-notification="showInfoNotification"
             @show-error-notification="showErrorNotification"
+            @show-warning-notification="showWarningNotification"
         />
     </component>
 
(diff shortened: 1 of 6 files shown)


nickpalladino (commented):
Should probably capture requirements somewhere for the changes here because it goes beyond what was specified in the issue. I'll defer to @DebWeigand on where makes the most sense.

DebWeigand (changes requested):
As per our discussion, I think it would be helpful to have the original message, as shown in the the Jira card BI-72 https://breedinginsight.atlassian.net/browse/BI-72, but highlighted in yellow as a warning "User created, but could not assign orcid. Orcid already in use.", with the form closed.

ctucker3:
> As per our discussion, I think it would be helpful to have the original message, as shown in the the Jira card BI-72 https://breedinginsight.atlassian.net/browse/BI-72, but highlighted in yellow as a warning "User created, but could not assign orcid. Orcid already in use.", with the form closed.

Pushed up changes. @DebWeigand to Nick's comment, do you have thoughts where to update requirements for the extra orcid duplicate check and the invalid orcid check?

nickpalladino (approved):
Approved assuming you and Deb will capture the requirements.

dmeidlin (approved):

DebWeigand (approved):

PR #50, opened 2020-10-19
[BI-626] Handle login error and logout user before starting login process

ctucker3:
Fixes issue with the nonce token (a token in the openid handshake that identifies a request upon return to open id requestor) expiring during the login process. The token still expires, but a bad token validation is handled and the user is notified on the front page to try again. I increase the expiration time to 10 minutes (it was 5 minutes). 

I also included a way to log out of open id before beginning the login process. This should be able to handle any Open ID provider if they have a logout URI. The logout url isn't required, and the process skips over that part if the logout attempt fails, or if the url isn't provided in the config. 

# Dependent PRs
https://github.com/Breeding-Insight/bi-api/pull/51

# Testing

1. Try changing the cookie expiration to 1s and check that you get the right page (see below)
2. Try logging in with an OrcID without a BI account, make sure you still get the right not authorized page. Make sure you can log into a good OrcID account after that. 
3. Try logging into an OrcID account, then log out, and then try logging in again. Make sure the orcid page pops up. 

# The way things should look

Here is the message the user gets when the authentication has timed out. 

screencapture-localhost-8080-2020-10-19-15_28_55

Here is a loading wheel I added to the OrcID button after you click it. I added this because we now have two calls in the login process, a BI backend health check and an orcid logout. If they take a few beats it lets the user know we are doing stuff. 
screencapture-localhost-8080-401-2020-10-19-16_33_04

Unchanged is our not authorized screen.  
screencapture-localhost-8080-2020-10-19-15_27_48

### DIFF
--- a/src/breeding-insight/dao/UserDAO.ts
+++ b/src/breeding-insight/dao/UserDAO.ts
@@ -118,7 +118,10 @@ export class UserDAO {
         }).catch((error) => reject(error));
 
     });
+  }
 
+  static openIdLogout(logoutUrl: string) {
+    return api.call({url: logoutUrl, method: 'get'});
   }
 
 }
--- a/src/views/Index.vue
+++ b/src/views/Index.vue
@@ -181,6 +181,12 @@
         Your session has expired. Please login again to continue.
       </h1>
       <h1
+        v-else-if="loginError"
+        class="is-size-5 has-text-primary"
+      >
+        An error has occurred during login. Please try again.
+      </h1>
+      <h1
         v-else
         class="is-size-5 has-text-primary"
       >
@@ -190,18 +196,19 @@
         To access to your breeding program, please log in.
       </p>
       <button
-        id="connect-orcid-button"
-        class="orcidBtn"
-        v-on:click="orcidLogin"
+          id="connect-orcid-button"
+          class="button orcidBtn"
+          v-bind:class="{'is-loading': loginProcessing, 'disabled': loginProcessing}"
+          v-on:click="orcidLogin"
       >
         SIGN IN with ORCID
         <img
-          id="orcid-id-icon"
-          src="https://orcid.org/sites/default/files/images/orcid_24x24.png"
-          width="24"
-          height="24"
-          class="is-pulled-right"
-          alt="ORCID iD icon"
+            id="orcid-id-icon"
+            src="https://orcid.org/sites/default/files/images/orcid_24x24.png"
+            width="24"
+            height="24"
+            class="is-pulled-right"
+            alt="ORCID iD icon"
         >
       </button>
       <p class="is-size-7 has-text-left">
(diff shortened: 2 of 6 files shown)


src/views/Index.vue
```diff
@@ -190,18 +196,19 @@
         To access to your breeding program, please log in.
       </p>
       <button
-        id="connect-orcid-button"
-        class="orcidBtn"
-        v-on:click="orcidLogin"
+          id="connect-orcid-button"
+          class="button orcidBtn"
+          v-bind:class="{'is-loading': loginProcessing, 'disabled': loginProcessing}"
+          v-on:click="orcidLogin"
```
timparsons, line 202: Can the button also be disabled when clicked to prevent multiple clicks by a user?

src/views/Index.vue
```diff
@@ -253,26 +261,41 @@
...
     mounted() {
-      if (this.loginRedirect || this.sessionExpired){
+      if (this.loginRedirect || this.sessionExpired || this.loginError){
         this.isLoginModalActive = true;
       }
     }
 
     // Methods
-    orcidLogin() {
+    async orcidLogin() {
```
timparsons, line 279: I assume this had to be changed to be asychronous because it would block and not allow the login button to show the spinner?

timparsons (changes requested):
Looks good!  Have one requested change for blocking additional clicks on the login button while it's in the loading mode.

eawoods:
Looked over design and workflow changes by request. The loading indicator is slick, nicely applied! All looks beautiful.

src/views/Index.vue
ctucker3: This is async now since I used `await` inside the method for the api calls instead of `.then().catch()`. It doesn't seem to effect Vue at all.

src/views/Index.vue
ctucker3: Yep! Removed the `disabled` class, and added a `disabled` conditional attribute.

timparsons (approved):

src/breeding-insight/dao/UserDAO.ts
```diff
@@ -118,7 +118,10 @@ export class UserDAO {
         }).catch((error) => reject(error));
 
     });
+  }
 
+  static openIdLogout(logoutUrl: string) {
+    return api.call({url: logoutUrl, method: 'get'});
```
nickpalladino, line 124: Are there any issues with doing this as an API call? I just remember one time we were doing something login/logout related as an api call rather than a browser redirect and it was causing us issues. Just wanted to make sure we don't run into that issue.

ctucker3: It seems to work fine. This is the way that they recommend it in the orcid documentation as well. 

https://members.orcid.org/api/resources/customize#logout

timparsons: I don't think the API can handle this because the logout function at ORCiD is most likely nuking all the cookies to log you out (as well as whatever they need to do on the backend).  Because of this, it needs to be from the client side (browser), and not from our API.

ctucker3: Yep, that was my understanding as well, and appears to be the case as I was testing it out during implementation. I played with the idea of adding a redirect to the orcid logout page with an ajax callback function (see link above) for a redirect back to bi-api as part of the backend login flow but this seemed like a more straight forward approach.

nickpalladino (approved):

PR #52, opened 2020-10-23
[BI-379] Display web and API build version on UI

timparsons:
# Overview
Display the current version of both bi-web and bi-api on the page, and provide a link to the GitHub release (or commit if running `develop`).  This can be useful for debugging purposes, as well as for identifying what features are deployed to a given server.  The implementation of updating and sharing version info of bi-api is in https://github.com/Breeding-Insight/bi-api/pull/54.  Here's a design of placement of the version info:

image

## Implementation
The version information for bi-web is stored in `package.json`.  The version name is stored in the `version` property, and a link to the release (or commit) is stored in a custom property named `versionInfo`.  These properties are automatically updated and committed to the repo via GitHub Actions.  

When merging to `master`, the action will use the output from release-drafter to set `version` to the name generated by release-drafter and `versionInfo` to a link to the eventual tag that will be created for the release.

When merging to `develop`, a different action will update the `version` property to be whatever the current version number is, and then append the build number of the action.  For example, if the version name is `v0.2.0`, and the action is running for the 10th time, then the resulting version name on the `develop` branch will be `v0.2.0+10`.  The `versionInfo` property will be set to the merge commit that triggered the action.

## Changes from design
The version info shows both the web and api versions separated by a forward slash.  For example: `web 0.2.0 / api 0.2.1`

## Requested Input

The design has the version information in the footer of the menu side panel.  However, if the user is not logged in, or is on a page that doesn't display a side panel, then they will not see any version information.  I took some liberty in seeing what it would look like if the version information was in the page footer.  If you could, please let me know your thoughts, and I'll adjust the display after collecting feedback:

Screen Shot 2020-10-27 at 5 55 25 PM

Screen Shot 2020-10-27 at 5 56 00 PM

### DIFF
--- a/src/components/layouts/BaseSideBarLayout.vue
+++ b/src/components/layouts/BaseSideBarLayout.vue
@@ -67,6 +67,11 @@
             <slot name="menu"></slot>
           </aside>
         </nav>
+        <div id="versionInfo" class="is-size-7 is-justify-content-center is-align-content-center is-flex">
+          <span class="is-centered">
+            <a :href="versionInfo" target="_blank">web {{versionName}}</a> / <a :href="apiVersionInfo" target="_blank">api {{ apiVersionName }}</a>
+          </span>
+        </div>
       </div>
 
 
@@ -97,6 +102,8 @@
   import {Component, Prop, Watch, Vue} from 'vue-property-decorator'
   import { MenuIcon } from 'vue-feather-icons'
   import {SandboxMode} from '@/util/config'
+  import * as api from "@/util/api";
+  import {ApiInfo} from "@/breeding-insight/model/ApiInfo";
 
 
   @Component( {
@@ -105,9 +112,14 @@
   export default class SideBarMaster extends Vue {
     sideMenuShownMobile: boolean = false;
     SandboxMode = SandboxMode;
+    apiInfo: ApiInfo = new ApiInfo("v0.0.0", "");
 
     @Prop()
     username!: string;
+
+    mounted() {
+      this.fetchApiVersion();
+    }
 
     @Watch('$route')
     onUrlChange(){
@@ -116,6 +128,31 @@
 
     get sandboxConfig() {
       return process.env.VUE_APP_SANDBOX;
+    }
+
+    get versionName() {
+      return process.env.VUE_APP_VERSION;
+    }
+
+    get versionInfo () {
+      return process.env.VUE_APP_VERSION_INFO;
+    }
+
+    get apiVersionName() {
+      return this.apiInfo.version;
+    }
+
+    get apiVersionInfo () {
+      return this.apiInfo.versionInfo;
+    }
+
+    private fetchApiVersion () {
+      api.call({
+        url: `${process.env.VUE_APP_BI_API_V1_PATH}/serverInfo`,
+        method: 'get'
+      }).then((response:any) => {
+        this.apiInfo = new ApiInfo(response.data.version, response.data.versionInfo);
+      })
     }
   }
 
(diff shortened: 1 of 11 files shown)


eawoods:
Placement thoughts: if the version info is going to be in the footer, it doesn't need to be in the sidebar. Also if in footer, move to the left so that it is associated more with the links than with the USDA and Cornell logos; it generates blank space that connects the version info to the logos rather than to the page itself.
Preferred option: move the version info into the lower left content area of pages that do not have a sidebar. In this way, version info will be in the same general location every time, for those users (likely to be mostly developers and QA) who are looking for it.

timparsons:
> Placement thoughts: if the version info is going to be in the footer, it doesn't need to be in the sidebar. Also if in footer, move to the left so that it is associated more with the links than with the USDA and Cornell logos; it generates blank space that connects the version info to the logos rather than to the page itself.
> Preferred option: move the version info into the lower left content area of pages that do not have a sidebar. In this way, version info will be in the same general location every time, for those users (likely to be mostly developers and QA) who are looking for it.

Thanks @eawoods!  Updated the display in the footer to be on the left, and added some logic to display in the footer when there's no sidebar (and visa versa)

eawoods (approved):
Works as expected, thanks for taking the note on left aligning the footer version.

src/components/layouts/BaseSideBarLayout.vue
```diff
@@ -117,6 +129,31 @@
...
+      return this.apiInfo.versionInfo;
+    }
+
+    private fetchApiVersion () {
+      api.call({
+        url: `${process.env.VUE_APP_BI_API_V1_PATH}/serverInfo`,
+        method: 'get'
+      }).then((response:any) => {
+        this.apiInfo = new ApiInfo(response.data.version, response.data.versionInfo);
+      })
```
nickpalladino, line 155: If it can't get version info from the api (missing version.properties or similar) will it just display v0.0.0?

src/components/layouts/Footer.vue
```diff
@@ -0,0 +1,125 @@
...
+
+  get apiVersionInfo () {
+    return this.apiInfo.versionInfo;
+  }
+
+  get showVersionInfo() {
+    return !this.$route.meta.layout || this.$route.meta.layout == "simple" || this.$route.meta.layout == "noSideBar";
+  }
+
+  fetchApiVersion () {
```
nickpalladino, line 115: Maybe this code could be moved to a service/dao and shared between the footer and sidebar?

src/components/layouts/Footer.vue
```diff
@@ -0,0 +1,125 @@
...
+import {Component, Prop, Watch, Vue} from 'vue-property-decorator'
+import * as api from "@/util/api";
+import {ApiInfo} from "@/breeding-insight/model/ApiInfo";
+
+@Component
+export default class Footer extends Vue {
+  private apiInfo: ApiInfo = new ApiInfo("v0.0.0", "");
+
+  mounted() {
+    this.fetchApiVersion();
```
nickpalladino, line 92: Haven't run it but is there any issue with the version flashing v0.0.0 to the actual version while the api call completes?

src/components/layouts/BaseSideBarLayout.vue
timparsons: Changed and pushed.

It will now display "unknown" if there is an error fetching the api version info.

src/components/layouts/Footer.vue
timparsons: Changed and pushed.

Created a `getApiInfo` method in `ServerManagementDAO`

src/components/layouts/Footer.vue
timparsons: Changed and pushed.

It will now display *loading* next to the api until the value is returned, and the Vue property is updated.

nickpalladino (approved):

dmeidlin (approved):

PR #55, opened 2020-11-11
[BI-601] - Implement role permissions on front end

ctucker3:
Introduces roles to the front end to hide/show different components. I followed rules set out on the card as well as defined rules from traits, which was not included in the cards. 

This will run separate from the bi-api branch of this. I still need to do some clean up on bi-api, but this branch shouldn't change during that clean up. 

I also modified the `ExpandableRowTable` element to have separate rules for archive and edit. They were a package deal before.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -18674,9 +18674,9 @@
       "integrity": "sha1-hnrHTjhkGHsdPUfZlqeOxciDB3c="
     },
     "typescript": {
-      "version": "3.5.3",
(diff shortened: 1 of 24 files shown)


timparsons (approved):
Casl looks great! Super clean and maintainable!

nickpalladino (approved):
The PR title id does not match the branch

ctucker3:
> The PR title id does not match the branch

Yeah the PR title is correct, the branch name is wrong. I realized after creating the PR that I was using a branch from a previous feature. It doesn't effect the code. Will it mess with your release packager?

PR #58, opened 2020-11-30
[BI-85,BI-646] - Create new trait UI

ctucker3:
Adds the ability to create a basic trait from the UI. 

Error display is not implemented yet, so you will get a generic error when trait creation fails. Open the developer console to see the errors printed out. You might need the console open before submitting for the errors to be printed. 

Some scale classes may fail because the specific scale additionals aren't in there yet. You will get errors like "Missing categories for scale class 'Categories'" or something along those lines. 

Some updates to the components. One to allow hiding of the label. One to allow the select field component to accept strings and objects.

### DIFF
--- a/src/breeding-insight/dao/ProgramDAO.ts
+++ b/src/breeding-insight/dao/ProgramDAO.ts
@@ -16,9 +16,10 @@
  */
 
 import {Program} from "@/breeding-insight/model/Program";
-import {BiResponse} from "@/breeding-insight/model/BiResponse";
+import {BiResponse, Response} from "@/breeding-insight/model/BiResponse";
 import * as api from "@/util/api";
 import {PaginationQuery} from "@/breeding-insight/model/PaginationQuery";
+
 
 export class ProgramDAO {
 
@@ -86,6 +87,15 @@ export class ProgramDAO {
       })
 
     }))
+  }
+
+  static async getObservationLevels(programId: string): Promise<BiResponse> {
+
+    const { data } =  await api.call({
+      url: `${process.env.VUE_APP_BI_API_V1_PATH}/programs/${programId}/observation_level`,
+      method: 'get'
+    }) as Response;
+    return new BiResponse(data);
   }
 
 }
--- a/src/breeding-insight/dao/ProgramDAO.ts
+++ b/src/breeding-insight/dao/ProgramDAO.ts
@@ -92,7 +92,7 @@ export class ProgramDAO {
   static async getObservationLevels(programId: string): Promise<BiResponse> {
 
     const { data } =  await api.call({
-      url: `${process.env.VUE_APP_BI_API_V1_PATH}/programs/${programId}/observation_level`,
+      url: `${process.env.VUE_APP_BI_API_V1_PATH}/programs/${programId}/observation-level`,
       method: 'get'
     }) as Response;
     return new BiResponse(data);
--- a/src/breeding-insight/dao/ProgramDAO.ts
+++ b/src/breeding-insight/dao/ProgramDAO.ts
@@ -92,7 +92,7 @@ export class ProgramDAO {
   static async getObservationLevels(programId: string): Promise<BiResponse> {
 
     const { data } =  await api.call({
-      url: `${process.env.VUE_APP_BI_API_V1_PATH}/programs/${programId}/observation-level`,
+      url: `${process.env.VUE_APP_BI_API_V1_PATH}/programs/${programId}/observation-levels`,
       method: 'get'
     }) as Response;
     return new BiResponse(data);
--- a/src/breeding-insight/model/Scale.ts
+++ b/src/breeding-insight/model/Scale.ts
@@ -18,7 +18,6 @@
 import {Category} from "@/breeding-insight/model/Category";
 
 export enum DataType {
-  Code = "CODE",
   Date = "DATE",
   Duration = "DURATION",
   Nominal = "NOMINAL",
--- a/src/breeding-insight/model/Scale.ts
+++ b/src/breeding-insight/model/Scale.ts
@@ -18,12 +18,12 @@
 import {Category} from "@/breeding-insight/model/Category";
 
 export enum DataType {
-  Date = "DATE",
-  Duration = "DURATION",
-  Nominal = "NOMINAL",
-  Numerical = "NUMERICAL",
-  Ordinal = "ORDINAL",
-  Text = "TEXT"
+  Date = "Date",
+  Duration = "Duration",
+  Nominal = "Nominal",
+  Numerical = "Numerical",
+  Ordinal = "Ordinal",
+  Text = "Text"
 }
 
 export class Scale {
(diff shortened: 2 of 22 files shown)


nickpalladino (changes requested):
Form looks good, on mobile too.

src/components/trait/forms/BaseTraitForm.vue
```diff
@@ -0,0 +1,150 @@
...
+    this.trait.method = new Method();
+    this.trait.scale = new Scale();
+  }
+
+  @Watch('trait', {deep: true})
+  emitTrait(val: Trait) {
+    this.$emit('trait-change', val);
+  }
+
+  filteredDataObj(data: string[]): string[] {
```
nickpalladino, line 112: May be a nice to have but this seems to match any letter in the observation level rather than a sequence starting with the first letter.

nickpalladino (changes requested):

src/breeding-insight/service/TraitService.ts
```diff
@@ -48,6 +54,8 @@ export class TraitService {
           let traits: Trait[] = [];
 
           if (biResponse.result.data) {
+            //TODO: Remove when backend default sorting is implemented
+            biResponse.result.data = PaginationController.mockSortRecords(biResponse.result.data);
```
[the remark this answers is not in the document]
ctucker3: I think this is needed for now to show the new trait at the top when it is added from the new trait from. Otherwise, it could appear on a different page and the user wouldn't see it.

src/components/trait/forms/BaseTraitForm.vue
ctucker3: Yeah it should match any substring with the options. So if typed, `lo` that should match `Plot` and `Lot' and 'Slot' and.... well I'll stop showing off my rhyming skills! When I'm on the interwebs, I usually like search bars like this instead of stricter search bars that search from the beginning.

src/breeding-insight/dao/ProgramDAO.ts
```diff
@@ -88,4 +89,13 @@ export class ProgramDAO {
     }))
   }
 
+  static async getObservationLevels(programId: string): Promise<BiResponse> {
+
+    const { data } =  await api.call({
+      url: `${process.env.VUE_APP_BI_API_V1_PATH}/programs/${programId}/observation_level`,
```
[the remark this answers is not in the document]
ctucker3: Yep, good catch. Changed and pushed.

src/breeding-insight/service/TraitService.ts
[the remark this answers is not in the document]
nickpalladino: Ah ya, that's right. Now you can do that on this page.

src/components/trait/forms/BaseTraitForm.vue
nickpalladino: ok, works for me.

src/breeding-insight/dao/ProgramDAO.ts
```diff
@@ -88,4 +89,13 @@ export class ProgramDAO {
     }))
   }
 
+  static async getObservationLevels(programId: string): Promise<BiResponse> {
+
+    const { data } =  await api.call({
+      url: `${process.env.VUE_APP_BI_API_V1_PATH}/programs/${programId}/observation-level`,
```
nickpalladino, line 95: can you add an 's' making it observation-levels

src/breeding-insight/model/Scale.ts
```diff
@@ -18,7 +18,6 @@
 import {Category} from "@/breeding-insight/model/Category";
 
 export enum DataType {
-  Code = "CODE",
```
nickpalladino, line 21: This is used in TraitDetailPanel, granted it's not really doing anything with it at this point.

nickpalladino (changes requested):

src/components/trait/TraitListsTable.vue
```diff
@@ -131,9 +180,14 @@ export default class TraitTable extends Vue {
   private collapseColumns = false;
   private StringFormatters = StringFormatters;
   private TraitStringFormatters = TraitStringFormatters;
+  private newTraitActive = false;
+  private methodClassOptions: string[] = Object.values(MethodClass);
+  private observationLevelOptions?: string[];
+  private scaleClassOptions: string[] = Object.values(DataType);
```
timparsons, line 186: Would it be possible to change the method and scale classes to be arrays of `option` -> `{id: <enum value>, name: <enum but with only first letter capitalized>}`?

timparsons (changes requested):
Looks great!!  I did have a couple of things:

1.  I accidentally ran into an issue where I clicked "Save" twice before the form closed (for whatever reason the app was being slow the first time I tried saving), and the trait table was updated.  After the table was updated, I had two records with duplicate information.  I verified this again, and was indeed able to create identical traits by double clicking the "Save" button.  @ctucker3 and I talked about this, and agreed that it would be best to disable the "Save" button after it's clicked to prevent double clicking, AND to add database constraints to check for uniqueness in the case that two identical trait POST requests are made simultaneously.

2. For the observation level input, I noticed that the autocomplete values are visible when focus is put on the input.  It wasn't obvious to me that a new observation level could be made by typing it into the input.  @eawoods could messaging be added to the input (or maybe even in the autocomplete box), and what should that look like?

src/breeding-insight/dao/ProgramDAO.ts
ctucker3: changed and pushed

eawoods:
> 2. For the observation level input, I noticed that the autocomplete values are visible when focus is put on the input.  It wasn't obvious to me that a new observation level could be made by typing it into the input.  @eawoods could messaging be added to the input (or maybe even in the autocomplete box), and what should that look like?

How about having placeholder text in the field that says "Start typing to see suggestions" ?

src/components/trait/TraitListsTable.vue
ctucker3: changed and pushed

src/breeding-insight/model/Scale.ts
ctucker3: removed

ctucker3:
> Looks great!! I did have a couple of things:
> 
> 1. I accidentally ran into an issue where I clicked "Save" twice before the form closed (for whatever reason the app was being slow the first time I tried saving), and the trait table was updated.  After the table was updated, I had two records with duplicate information.  I verified this again, and was indeed able to create identical traits by double clicking the "Save" button.  @ctucker3 and I talked about this, and agreed that it would be best to disable the "Save" button after it's clicked to prevent double clicking, AND to add database constraints to check for uniqueness in the case that two identical trait POST requests are made simultaneously.

My database constraints didn't work with the double click, although they would work with a regular query. I'm thinking those queries are being executed simultaneously so the duplicates aren't being caught. I created a new card to look into this:

https://breedinginsight.atlassian.net/browse/BI-689

The prevention of the double click on the UI is implemented for traits. It needs to be added to the other forms still though. 

> 2. For the observation level input, I noticed that the autocomplete values are visible when focus is put on the input.  It wasn't obvious to me that a new observation level could be made by typing it into the input.  @eawoods could messaging be added to the input (or maybe even in the autocomplete box), and what should that look like?

Implemented Liz's suggestion of the placeholder.

timparsons (approved):

nickpalladino (approved):

PR #60, opened 2020-12-02
[BI-671,BI-679] - Create new trait - Scale class 'ordinal scale' and Scale class 'nominal class'

ctucker3:


src/components/trait/forms/BaseTraitForm.vue
```diff
@@ -50,6 +50,31 @@
...
+        <DateTraitForm />
+      </template>
+      <template v-if="trait.scale && trait.scale.dataType === DataType.Duration">
+        <DurationTraitForm />
+      </template>
+      <template v-if="trait.scale && trait.scale.dataType === DataType.Nominal">
+        <CategoryTraitForm
+            v-on:update="trait.scale.categories = $event"
+            v-bind:type="DataType.Nominal"
+        />
```
timparsons, line 72: This element and `Ordinal` could be combined.  The `type` passed to `CategoryTraitForm` can be the value of `trait.scale.dataType`

timparsons (changes requested):

src/components/trait/forms/BaseTraitForm.vue
```diff
@@ -50,6 +50,31 @@
       </div>
 
       <!-- Scale options -->
+      <template v-if="trait.scale && trait.scale.dataType === DataType.Ordinal">
+        <CategoryTraitForm
+          v-on:update="trait.scale.categories = $event"
+          v-bind:type="DataType.Ordinal"
+        />
+      </template>
+      <template v-if="trait.scale && trait.scale.dataType === DataType.Text">
```
nickpalladino, line 59: There's also a method to do a case insensitive comparison for the DataType in the Scale class. Not important here since it's a new trait but may be for edit where that info could possibly come in differently.

nickpalladino (approved):

src/components/trait/forms/BaseTraitForm.vue
ctucker3: Ok, that might be good to check. I was thinking that those data types got transformed into the enum values which is why I didn't worry about case matching. I'll leave that for the edit form card if that is cool with you?

src/components/trait/forms/BaseTraitForm.vue
ctucker3: Good catch. This is changed in one of the next PR's I have open.

timparsons (approved):

PR #62, opened 2020-12-08
[BI-684, BI-682] - Calendar scale and Computation method class

ctucker3:
Added the calendar scale class and the computation method class. Also added a history to the scale classes and the method formula.

### DIFF
--- a/src/components/trait/forms/BaseTraitForm.vue
+++ b/src/components/trait/forms/BaseTraitForm.vue
@@ -212,13 +212,37 @@ export default class TraitTable extends Vue {
 
     // Save history of current scale class
     if (this.trait.scale!.dataType) {
-      this.scaleHistory[this.trait.scale!.dataType] = {...this.trait.scale!};
+      // Nominal and ordinal save histories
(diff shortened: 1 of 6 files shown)


timparsons (approved):

src/components/trait/forms/BaseTraitForm.vue
```diff
@@ -173,38 +180,55 @@ export default class TraitTable extends Vue {
...
+    // Computation method class is always numerical
+    if (this.trait!.method!.methodClass === MethodClass.Computation) {
+      this.setScaleClass(DataType.Numerical);
     }
-    // Switching for ordinal or nominal to something else
-    else {
-      this.trait.scale!.categories = undefined;
+  }
+
+  setScaleClass(value: string) {
```
nickpalladino, line 211: Is the state of ordinal and nominal supposed to be shared with each other? Looks like now they each keep their own state.

ctucker3: I left it out because I wanted to limit the logic on the UI and it would be one more thing to maintain. But, it is part of the requirements, so I added it in. Thanks for the catch!

nickpalladino (approved):

PR #63, opened 2020-12-14
[BI-685] - Map backend trait validation errors to front end

ctucker3:
PAIRED WITH TEMPLATE UPDATE. UPDATE LINK AFTER APPROVAL:

https://github.com/Breeding-Insight/requirements/pull/16

Also includes a fix to feather icons so that they line up with text properly. 

Maps the server errors to the UI inputs. Some custom parsing needed to be done for `units` because the field they go into is `scaleName`. It's not perfect, but it works for what we are doing. 

The design for the error icons changed slightly after speaking with Liz. I put them in the error text instead of in the input because it was easier for the same effect. 

screencapture-localhost-8080-programs-d111dfa6-3acd-4906-9aa5-c8f3f4fffce9-traits-list-2020-12-14-15_27_21

### DIFF
--- a/src/assets/scss/main.scss
+++ b/src/assets/scss/main.scss
@@ -464,9 +464,3 @@ div.sentence-input {
   fill: none;
   stroke-width: 2;
 }
-
-// Fix vie feather icons
-svg.feather {
-  vertical-align: middle;
-  margin-top: -4px;
-}
--- a/src/assets/scss/main.scss
+++ b/src/assets/scss/main.scss
@@ -464,3 +464,9 @@ div.sentence-input {
   fill: none;
   stroke-width: 2;
 }
+
+// Fix vie feather icons
+svg.feather {
+  vertical-align: middle;
+  margin-top: -4px;
+}
--- a/src/assets/scss/main.scss
+++ b/src/assets/scss/main.scss
@@ -184,6 +184,12 @@ footer {
     & .field span.form-error {
       display: block;
     }
+    & input {
+      border-color: $danger;
+    }
+    & select {
+      border-color: $danger;
+    }
   }
   &:not(.field--error) {
     & .field span.form-error {
--- a/src/breeding-insight/model/errors/ValidationError.ts
+++ b/src/breeding-insight/model/errors/ValidationError.ts
@@ -16,11 +16,37 @@
  */
 
 import {RowError} from "@/breeding-insight/model/errors/RowError";
+import {FieldError} from "@/breeding-insight/model/errors/FieldError";
 
 export class ValidationError {
-  rows: RowError[];
+  rows?: RowError[];
 
-  constructor(rows: RowError[]){
-    this.rows = rows;
+  constructor(rows?: RowError[]){
+    if (rows) {
+      this.rows = rows.map(row => new RowError(row.rowIndex, row.errors));
+    }
+  }
+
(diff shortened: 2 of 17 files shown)


timparsons (changes requested):

src/breeding-insight/model/errors/ValidationError.ts
```diff
@@ -16,11 +16,51 @@
...
+    }
+  }
+
+  condenseErrorsSingleRow() {
+    let errorSentence: string = '';
+    if (this.rows) {
+      for (const row of this.rows) {
+        if (row.errors) {
+          for (const field of row.errors) {
+            errorSentence += `${field.field}: ${field.errorMessage}; `;
```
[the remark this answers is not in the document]
ctucker3: That's a good point, I can remove the fields for the error sentence. The other fields don't have the error sentence yet, but I think they can just use the errorMessage as well since we will have to custom create those sentences from the UI validation.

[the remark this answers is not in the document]
ctucker3: Changed and pushed

timparsons (approved):

src/components/trait/TraitListsTable.vue
```diff
@@ -213,20 +214,19 @@ export default class TraitTable extends Vue {
...
       this.newTraitActive = false;
     } catch (error) {
-      // TODO: Pass errors to the new data form
       if (error instanceof ValidationError) {
-        const fieldErrors: FieldError[] = error.rows[0].errors;
-        for (const fieldError of fieldErrors) {
-          this.$log.error(fieldError);
-        }
+        this.validationHandler = error;
+        this.$emit('show-error-notification', `Error creating trait. ${this.validationHandler.condenseErrorsSingleRow()}`);
```
nickpalladino, line 226: Not exactly a 1-to-1 mapping between error message and UI but probably no getting around that. For example, if you don't change the form from the loading state and try to submit, one of the error messages will be missing scale name which isn't present on the form and would be rectified by selecting a scale class/setting unit.

src/assets/scss/main.scss
```diff
@@ -458,3 +464,9 @@ div.sentence-input {
   fill: none;
   stroke-width: 2;
 }
+
+// Fix vie feather icons
+svg.feather {
```
nickpalladino, line 469: Since this is a global change, it affects alignment of icons used elsewhere like the table row chevrons, and possibly other stuff. I preferred the old alignment of chevrons, didn't check everywhere else.

nickpalladino (commented):
One thing that may be weird for people is the way our front end validations clear prior to submit where the backend ones don't until a re-save is attempted. Maybe we should change how the front end ones work for consistency at some point?

src/components/trait/TraitListsTable.vue
ctucker3: Good catch, I mapped the scale name error for the fields but not for the top notification. I added mappings for those, changes are pushed up. The scale name is the tricky part because it has multiple roles depending on the data type.

src/assets/scss/main.scss
ctucker3: Yeah, I liked the old alignment for the chevrons too. I missed that when checking the icon changes. I changed it back and made appropriately changes to my new warning labels directly.

nickpalladino (approved):

PR #64, opened 2020-12-16
Feature/bi625

dmeidlin:
The domain for bi-api-root was from localhost:8081 to localhost since biapi will not have a port exposed on teh host.

Fixed a bug in the docker-copmse.yml files so that the source code and tests are now copied into biweb when in production mode.

### DIFF
--- a/.env.development
+++ b/.env.development
@@ -5,7 +5,7 @@ VUE_APP_OPENID_LOGOUT_URL=https://sandbox.orcid.org/userStatus.json?logUserOut=t
 VUE_APP_SANDBOX=public
 
 # Only edit this one
-VUE_APP_BI_API_ROOT=http://localhost:8081
+VUE_APP_BI_API_ROOT=http://localhost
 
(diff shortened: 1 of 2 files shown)


ctucker3 (approved):
I'm a little confused how `http://localhost` gets routed to bi-api. But, I'm guessing you tried it out and it worked and biproxy is able to handle it, so I'm good with it.

nickpalladino (approved):

dmeidlin:
> I'm a little confused how `http://localhost` gets routed to bi-api. But, I'm guessing you tried it out and it worked and biproxy is able to handle it, so I'm good with it.

Ports 80 and 443 for localhost are connected to the reverse proxy.  The rule set for the reverse proxy determines which requests are sent to a specific upstream server: in the present case that's either bi-web or bi-waf(which sends requests along to bi-api). Requests sent to /server-info and the /sso and /v1 namespaces are sent to bi-waf; all others are sent to bi-web.

PR #66, opened 2021-01-05
[BI-86] - Edit trait form

ctucker3:
Adds the ability to edit traits on the trait list page. The side panel table was getting messy trying to manage all of the states with the edit panel added, so I changed it to have a state manager and handles page interactions through events. Some small changes might need to be made once the PUT traits endpoint is up and running. 

I need to fix some tests, but other than that is ready for review.


nickpalladino (commented):
When I open the edit form for a trait and don't edit anything it will let me click on another row and bring that edit form up. If I still don't change anything and click another row, a dialog will pop up asking me if I want to close the panel. I would think it should just allow me to keep selecting new rows without showing that if I haven't edited anything.

nickpalladino (commented):
Was there a decision on the form closing (wither save or cancel) closing the side panel vs returning to the side panel details which is where you navigated to the edit form from. The second option seems more intuitive to me.

nickpalladino (commented):
Missing chevron that was there before on selected row and different behavior of Show details:
image

nickpalladino (commented):
Design reference shows primary-light highlight for row selection when editing, same as in expandable row table and also shows a chevron on the selected row.

nickpalladino (commented):
I see edit is only on the traits in the traits list page and not import curation, do we have a separate card for that? Can't remember what we were doing with that?

nickpalladino (commented):
The breakpoint column collapse doesn't seem to work anymore. Used to collapse down to just showing name column on small screens, now doesn't.

ctucker3:
> The breakpoint column collapse doesn't seem to work anymore. Used to collapse down to just showing name column on small screens, now doesn't.

@timparsons can you try this on yours to see if it works? I tested this on the same browser as Nick is using and the column collapsing still worked, so I am wondering if it is his computer, not the browser. Nick uses Mozilla version 84. If it works for you, I think Nick and I will have to pair program to get this working.

ctucker3:
> I see edit is only on the traits in the traits list page and not import curation, do we have a separate card for that? Can't remember what we were doing with that?

That is in a separate card:

https://breedinginsight.atlassian.net/browse/BI-586

src/breeding-insight/model/Method.ts
```diff
@@ -36,7 +36,19 @@ export class Method {
...
+  }
+
   static methodClassEquals(classString: string, type: MethodClass): boolean {
     return classString.toUpperCase() === type.toUpperCase();
   }
+
+  equals(method?: Method): boolean {
+    if (!method) {return false;}
+    return (this.methodName === method.methodName) &&
+      (this.methodClass === method.methodClass) &&
```
nickpalladino, line 50: Will there be any possible issues with case sensitivity on this, didn't follow it all the way through? May be more relevant in future if interacting directly with BrAPI services.

nickpalladino (approved):

src/components/VBreakpoint.ts
```diff
@@ -20,7 +20,7 @@ import { extend } from 'vue-breakpoint-component'
 const config = {
   // breakpoints for bulma
   breakpoints: {
-    mobile: '(min-width: 529px)',
+    mobile: '(min-width: 0px)',
```
timparsons, line 23: why was this changed to zero?

timparsons (approved):
Looks good! Had one question on a change in the breakpoint for mobile.

src/components/VBreakpoint.ts
ctucker3: There is nothing below tablet but mobile, so I just this to zero so that this the bottom most default. It wouldn't switch to mobile before if you had a screen size smaller than 529px.

src/breeding-insight/model/Method.ts
ctucker3: I think it makes sense to make this method check case insensitive. For other equality checks like trait name, I would think we want case sensitive matching so that if some changes `Plant Height` to `plant height` we will recognize that as a change. Caps to lower cases might be important for some fields. 

I'll make the method check case insensitive before merging.

src/components/VBreakpoint.ts
timparsons: 👍

PR #69, opened 2021-01-27
[BI-695]  ISSUE: Multiple clicks on Save results in equally multiple entities created for Programs and Locations

timparsons:
`NewDataForm` has been updated to automatically disable the "Save" button once it's been clicked to prevent double clicking.  The button will also show a loading indicator, to indicate that something is happening.

I also refactored the structure of the "views" folder a little bit to group views together based on their function.

Lastly, I added a "config" node in `package.json` and moved custom variables used for building and startup as I was having issues getting bi-web to start up.  This has no external affect on running any of the npm scripts for bi-web.


src/components/forms/NewDataForm.vue
```diff
@@ -72,19 +74,27 @@
       return this.$v.newRecord;
     }
 
+    disableSaveBtn() {
```
nickpalladino, line 77: I see the saveBtnActive prop is being used by parent components such as TraitsListTable to disable it when being saved. Just wondering why we have that as well as this saveStarted flag. I see there is that validation code prior to the events notifying the parent, not sure if that was an issue before when the original fix was put in? Also don't see programs or locations tables using saveBtnActive.

timparsons: Good catch.  I think the logic for deactivating the "Save" button should be in the *DataForm (either `NewDataForm` or `EditDataForm`), and not in the parent component.  I need to look through things again, but I thought I saw a place where the "Save" button was deactivated when opening the form, and that's why I didn't remove the prop.  Also need to update the `EditDataForm` to have the deactivate logic.

timparsons:
Realized that I missed handling edits after Nick had done a review.  Most recent commit accounts for both new data forms and edit forms.  Approach implemented was to create a `EventBus` to signaling when a save activity is starting, and when it is complete (complete means it was either successful or not).  Each parent component was updated to now have an instance of `DataFormEventBusHandler`, and the parent is responsible for signaling when the save activity is complete.

src/components/forms/DataForm.vue
```diff
@@ -0,0 +1,115 @@
...
+      return;
+    } else {
+      // debouncing
+      setTimeout(() => {
+        this.$emit('submit');
+
+        if (this.$v.record) {
+          this.$v.record.$reset();
+        }
+      }, 250);
```
ctucker3, line 102: What is the purpose of this?

ctucker3 (approved):
Seems to work fine when I was poking around! A couple of comments and questions, but nothing to hold back approval.

package.json
```diff
@@ -73,7 +77,5 @@
     "vue-cli-plugin-axios": "0.0.4",
     "vue-template-compiler": "^2.6.10"
   },
-  "devport": "8080",
-  "task_path": "./task",
   "versionInfo": "https://github.com/Breeding-Insight/bi-web/commit/b06cac8e7d9d8ec468594add7b9e543f9252ff29"
```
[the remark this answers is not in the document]
timparsons: `devport` and `task_path` are both used to run npm tasks, where `versionInfo` is informative just as `version` is.

src/components/forms/DataFormEventBusHandler.ts
```diff
@@ -0,0 +1,43 @@
...
+import {EventStore} from "@/util/EventStore";
+
+export class DataFormEventBusHandler {
+  bus = new Vue();
+
+  //Events
+  static readonly SAVE_STARTED_EVENT = "save-activity-started"
+  static readonly SAVE_COMPLETE_EVENT = "save-activity-complete"
+
+  public readonly id = Math.random();
```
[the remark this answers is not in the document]
timparsons: Good catch! That was left over from some debugging I was doing.  I'll get this removed.

src/components/forms/DataForm.vue
timparsons: I've updated this to be a more proper debouncer in that every time the "Save" button is pressed, it resets the timeout.  This means that the save request won't be initiated until there's been 250ms since the last "Save" button press, which is also enough time for Vue to go through its lifecycle to update the UI to disable it and prevent double clicking.

src/components/forms/DataFormEventBusHandler.ts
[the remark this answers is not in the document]
timparsons: Removed and pushed

nickpalladino (approved):

DebWeigand (approved):

PR #70, opened 2021-01-29
[BI-563] Archive Individual Trait

dmeidlin:
The Trait model was updated to include the "active" property and set it to true by default.

The Archiving features in the UI can be activated/deactivated by toggle the state of the archivable prop in TraitsListTable.
Archivable should be set to false until the "active" property is included in trait data from bi-api.

Given: The user has navigated to the "Trait Management" screen
And: the user has selected a trait to view
When: the user clicks "Archive" on the detail pane
Then: ensure that a confirmation dialog appears

Given: The user has clicked “Archive”
When: they confirm archiving the trait
Then: ensure that the trait detail pane is closed
And: ensure that the row shows an indicator that the trait is archived
And: ensure that a success notification is displayed at the top of the screen

Given: The user has clicked “Archive”
When: they decline archiving the trait
Then: ensure that the archive confirmation dialog closes
And: ensure that the trait is not archived

Given: The user has navigated to the "Trait Management" screen
When: the user has selected an archived trait to view
Then: ensure that the ‘Archived’ tag is displayed at the bottom of the detail pane

Given: that there is more than one page of traits in the trait list
When: the user archives a trait that is not on the first page
Then: ensure that the page in view is reset to the first page

### DIFF
--- a/package.json
+++ b/package.json
@@ -13,7 +13,7 @@
     "test:unit": "vue-cli-service test:unit"
   },
   "config": {
-    "devport": "8080",
+    "devport": "8082",
     "task_path": "./task"
   },
   "dependencies": {
--- a/package.json
+++ b/package.json
@@ -13,7 +13,7 @@
     "test:unit": "vue-cli-service test:unit"
   },
   "config": {
-    "devport": "8082",
+    "devport": "8080",
     "task_path": "./task"
   },
   "dependencies": {
--- a/src/components/trait/TraitDetailPanel.vue
+++ b/src/components/trait/TraitDetailPanel.vue
@@ -70,6 +70,16 @@
       <p class="has-text-weight-bold mt-3 mb-0">Description of collection method</p>
       <p>{{data.method.description}}</p>
 
+      <template v-if="!archivable">
+        <p class="has-text-weight-bold mt-3 mb-0">Included in Favorites</p>
+        <b-button
+            type="is-light"
+            style="background: lightgray"
+            v-if="!archivable">
+          Archived
+        </b-button>
+      </template>
+
       <!-- maybe break out controls for reuse eventually -->
       <div class="columns is-mobile is-centered pt-6">
         <div class="column is-narrow">
--- a/src/components/trait/TraitDetailPanel.vue
+++ b/src/components/trait/TraitDetailPanel.vue
@@ -76,7 +76,7 @@
             size="is-small"
             style="background: lightgray"
             class="archive-tag"
-            v-if="!data.active && archivable">
+            v-if="!data.active">
           Archived
         </b-button>
       </template>
@@ -95,7 +95,7 @@
         </div>
         <div class="column is-narrow">
           <a
-            v-if="archivable && data.active"
+            v-if="data.active"
             v-on:click="$emit('archive', data)"
             v-on:keypress.enter.space="$emit('archive', data)"
             tabindex="0"
--- a/src/components/trait/TraitDetailPanel.vue
+++ b/src/components/trait/TraitDetailPanel.vue
@@ -76,7 +76,7 @@
             size="is-small"
             style="background: lightgray"
             class="archive-tag"
-            v-if="!data.active">
+            v-if="!data.active && archivable">
           Archived
         </b-button>
       </template>
--- a/src/components/trait/TraitDetailPanel.vue
+++ b/src/components/trait/TraitDetailPanel.vue
@@ -73,9 +73,10 @@
       <template v-if="!archivable">
         <p class="has-text-weight-bold mt-3 mb-0">Included in Favorites</p>
         <b-button
-            type="is-light"
+            size="is-small"
             style="background: lightgray"
-            v-if="!archivable">
+            class="archive-tag"
+            v-if="!data.active">
           Archived
         </b-button>
       </template>
--- a/src/components/trait/TraitDetailPanel.vue
+++ b/src/components/trait/TraitDetailPanel.vue
@@ -91,6 +91,14 @@
             >
             Archive
           </a>
+          <a
+              v-else
+              v-on:click="$emit('restore')"
+              v-on:keypress.enter.space="$emit('restore')"
+              tabindex="0"
+          >
+            Restore/Unarchive
+          </a>
         </div>
       </div>
     </template>
--- a/src/components/trait/TraitDetailPanel.vue
+++ b/src/components/trait/TraitDetailPanel.vue
@@ -70,7 +70,7 @@
       <p class="has-text-weight-bold mt-3 mb-0">Description of collection method</p>
       <p>{{data.method.description}}</p>
 
-      <template v-if="!archivable">
+      <template v-if="archivable && !data.active">
         <p class="has-text-weight-bold mt-3 mb-0">Included in Favorites</p>
         <b-button
             size="is-small"
@@ -95,18 +95,18 @@
         </div>
         <div class="column is-narrow">
           <a
-            v-if="archivable"
+            v-if="archivable && data.active"
             v-on:click="$emit('archive', data)"
-            v-on:keypress.enter.space="$emit('archive')"
+            v-on:keypress.enter.space="$emit('archive', data)"
             tabindex="0"
             >
             Archive
           </a>
           <a
-              v-else
-              v-on:click="$emit('restore')"
-              v-on:keypress.enter.space="$emit('restore')"
-              tabindex="0"
+            v-if="archivable && !data.active"
+            v-on:click="$emit('restore', data)"
+            v-on:keypress.enter.space="$emit('restore', data)"
+            tabindex="0"
           >
             Restore/Unarchive
           </a>
--- a/src/components/trait/TraitDetailPanel.vue
+++ b/src/components/trait/TraitDetailPanel.vue
@@ -70,7 +70,7 @@
       <p class="has-text-weight-bold mt-3 mb-0">Description of collection method</p>
       <p>{{data.method.description}}</p>
 
-      <template v-if="archivable && !data.active">
+      <template v-if="!data.active">
         <p class="has-text-weight-bold mt-3 mb-0">Included in Favorites</p>
         <b-button
             size="is-small"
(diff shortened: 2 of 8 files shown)


package.json
```diff
@@ -73,7 +73,7 @@
     "vue-cli-plugin-axios": "0.0.4",
     "vue-template-compiler": "^2.6.10"
   },
-  "devport": "8080",
+  "devport": "8082",
```
ctucker3, line 76: Does this need to stay changed or can we change it back to 8080?

src/components/trait/TraitDetailPanel.vue
```diff
@@ -70,6 +70,17 @@
       <p class="has-text-weight-bold mt-3 mb-0">Description of collection method</p>
       <p>{{data.method.description}}</p>
 
+      <template v-if="archivable && !data.active">
```
ctucker3, line 73: Can this just be `!data.active`? The data can be archived without the table providing archiving abilities. One route I can think of is if the status is set to inactive in the import (not sure if we support that yet).

src/components/trait/TraitListsTable.vue
```diff
@@ -29,8 +29,18 @@
       </section>
       <div class="columns">
         <div class="column is-whole has-text-centered buttons">
-          <button v-on:click="modalDeleteHandler()" class="button is-danger"><strong>Yes, remove</strong></button>
-          <button v-on:click="deactivateActive = false" class="button">Cancel</button>
+          <button
+            class="button is-danger"
+            v-on:click="modalDeleteHandler"
+          >
+            <strong>Yes, {{ focusTrait.active ? 'remove' : 'restore' }}</strong>
```
ctucker3, line 36: Should this be `Yes, archive` instead of `Yes, remove`?

src/components/trait/TraitListsTable.vue
```diff
@@ -245,6 +267,30 @@ export default class TraitTable extends Vue {
...
+    } else {
+      this.deactivateWarningTitle = `Restore "${focusTrait.traitName}" to ${this.activeProgram!.name!}?`;
+    }
+    this.focusTrait = focusTrait;
+    this.deactivateActive = true;
+  }
+
+  async modalDeleteHandler(){
+    try {
+      const traitClone = JSON.parse(JSON.stringify(this.focusTrait));
```
ctucker3, line 282: Never knew that was a way to clone! There is also the `Trait.assign` function in the trait object if you want a trait class instance.

ctucker3 (approved):
Looks good! I didn't run it locally but the code looks good. I had a few questions and comments, but approving since I'll be out the rest of the week.

package.json
nickpalladino: +1

src/components/trait/TraitListsTable.vue
```diff
@@ -117,10 +134,13 @@
           v-bind:edit-active="traitSidePanelState.editActive"
           v-bind:edit-btn-active="editFormBtnActive"
           v-bind:editable="true"
+          v-bind:archivable="false"
```
nickpalladino, line 137: Why is this false? I guess that's why I don't see an archive link in the detail pane?

src/components/trait/TraitListsTable.vue
nickpalladino: +1

src/components/trait/TraitListsTable.vue
```diff
@@ -245,6 +267,30 @@ export default class TraitTable extends Vue {
     });
   }
 
+  activateArchive(focusTrait: Trait){
+    if (focusTrait.active) {
+      this.deactivateWarningTitle = `Remove "${focusTrait.traitName}" from ${this.activeProgram!.name!}?`;
```
nickpalladino, line 272: Similar to previous comment, maybe this should also use archive terminology

package.json
dmeidlin: The webserver is running on 8082 because the brapi-server is running on 8080.
There are other PRs related to this pending review:
https://github.com/Breeding-Insight/bi-docker-stack/pull/8,
https://github.com/Breeding-Insight/bi-api/pull/67

src/components/trait/TraitListsTable.vue
dmeidlin: Assign would probably work in this case, but I generally stay away from it because it can cause errors if a property of interest is deeply nested (see https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/assign).

Also, I remember seeing a comparison of computational effort required for the JSON parsing method for deep cloning versus using methods in underscore or other libraries, and JSON parsing came out on top, so I generally use it for cloning.

src/components/trait/TraitListsTable.vue
dmeidlin: Just documenting what we talked about earlier:
The frontend logic works once the "active" trait is returned from the api. Mocking out flagging some traits as archived and others not while preventing the mocks to be overwritten by calls to the trait api was beyond the scope of this work and not really of great value. Without the archivable boolean flag stored per trait the feature can't function properly, so the archivable flag is set to false to turn the feature off until the backend is ready.

src/components/trait/TraitListsTable.vue
dmeidlin: I used "remove" because it was already in place in the warning modal. The following is from commit ee46948:
`<template>
  <section id="traitTableLabel">
    <WarningModal
      v-bind:active.sync="deactivateActive"
      v-bind:msg-title="'Remove trait from Program name?'"
      v-on:deactivate="deactivateActive = false"
    >
      <section>
        <p class="has-text-dark">
          Program-related data referencing this trait will not be affected by this change.
        </p>
      </section>
      <div class="columns">
        <div class="column is-whole has-text-centered buttons">
          <button v-on:click="modalDeleteHandler()" class="button is-danger"><strong>Yes, remove</strong></button>
          <button v-on:click="deactivateActive = false" class="button">Cancel</button>
        </div>
      </div>              
    </WarningModal>
`
The design reference doesn't show anything for the warning modal, and I think "archive" is more accurate than "remove", so I'm fine with making the change.

dmeidlin: Sorry the code didn't typeset well; try again...
   <WarningModal
      v-bind:active.sync="deactivateActive"
      v-bind:msg-title="'Remove trait from Program name?'"
      v-on:deactivate="deactivateActive = false"
    >
      <section>
        <p class="has-text-dark">
          Program-related data referencing this trait will not be affected by this change.
        </p>
      </section>
      <div class="columns">
        <div class="column is-whole has-text-centered buttons">
          <button v-on:click="modalDeleteHandler()" class="button is-danger"><strong>Yes, remove</strong></button>
          <button v-on:click="deactivateActive = false" class="button">Cancel</button>
        </div>
      </div>              
    </WarningModal>

dmeidlin: Ok, this works. Sorry for spamming your inbox. :-)
```
   <WarningModal
      v-bind:active.sync="deactivateActive"
      v-bind:msg-title="'Remove trait from Program name?'"
      v-on:deactivate="deactivateActive = false"
    >
      <section>
        <p class="has-text-dark">
          Program-related data referencing this trait will not be affected by this change.
        </p>
      </section>
      <div class="columns">
        <div class="column is-whole has-text-centered buttons">
          <button v-on:click="modalDeleteHandler()" class="button is-danger"><strong>Yes, remove</strong></button>
          <button v-on:click="deactivateActive = false" class="button">Cancel</button>
        </div>
      </div>              
    </WarningModal>
```

src/components/trait/TraitListsTable.vue
dmeidlin: Agreed.

src/components/trait/TraitDetailPanel.vue
dmeidlin: Yes, the !data.active check is all that is required once the backend returns the active status as part of the trait data.

When the traits are pulled from the backend the trait data is looped over all defined key values when creating the trait objects for the frontend. Since the active property does not appear in the trait data, data.inactive will be undefined and !data.active will be true for all traits even though they are active. So the achivable check is in place to turn these sections off until the backend can support archiving.  Once the backend part is working the archivable check can be removed.

src/components/trait/TraitListsTable.vue
nickpalladino: Can we fix bi-api as part of this card so that we can test this feature? It looks like createTraits in TraitService needs to set the active field in the TraitEntity as it's writing null to the database currently.

package.json
ctucker3: The brapi server runs internally in the container as 8080 and externally as 8083, so I think the web could stay as 8080.

src/components/trait/TraitListsTable.vue
ctucker3: That makes sense. Just to clarify, the `Trait.assign` is a custom function of ours that is implemented in our classes.

ctucker3 (changes requested):
From our discussion during sprint planning. Can we move the `Archive` tag to another column or create a new card to do that?

src/components/trait/TraitListsTable.vue
ctucker3: For traceability, I fixed bi-api in the BI-732 PR that is currently open.

DebWeigand (changes requested):
This scenario, as is written in the card, does not pass (unless I am misunderstanding)
**Given:** that there is more than one page of traits in the trait list
**When:** the user archives a trait that is not on the first page
**Then:** ensure that the page in view is reset to the first page

after archiving the trait on the second page, the view does not go back to the first page. BUT, do we really want it too?  Don't we want to see it in the list as archived on the original page it was on?  Or, I may be just misunderstanding...

package.json
dmeidlin: Ok, will change the port back.

dmeidlin:
@ctucker3 I think it's fine adding the archive badge as another column in this PR. Now that the backend is working I'll merge in the updates from https://github.com/Breeding-Insight/bi-web/pull/77 and add the change with the extra column.

ctucker3 (approved):

DebWeigand (approved):

nickpalladino (approved):

PR #71, opened 2021-02-01
[BI-697] Orcid Icon missing

dmeidlin:
Description: ORCID Icon missing
Component/Version: all
Steps to Reproduce:

use any BI instance

click login

Expected result:
see Welcome to Breeding Insight pop-up with 2 ORCID Icons within

Actual Result:
broken links for the ORCID Icons

Simply replacing the current orcid remote url with a path to a BI static asset has the downside that updating the asset when orcid changes their icon is required.  This fix keeps the remote url as the src but uses Vue to dynamically change the src to point to the BI asset if the image is broken.

NOTE: The only tricky part about dynamically handling relative paths to assets in Vue has to do with webpack. It's necessary to wrap require('@/path/to/asset') around the path so that webpack loads the image via JS and can read the path correctly:https://github.com/vuejs-templates/webpack/issues/450.


timparsons (approved):
Great approach!

nickpalladino (approved):

DebWeigand (approved):

PR #72, opened 2021-02-02
[BI-677] - Connect PUT traits to back end

ctucker3:
Hook up the PUT traits on the front end to the backend. 

See other PRs:
https://github.com/Breeding-Insight/bi-api/pull/68
https://github.com/Breeding-Insight/sgn/pull/7

### DIFF
--- a/src/breeding-insight/dao/TraitDAO.ts
+++ b/src/breeding-insight/dao/TraitDAO.ts
@@ -48,12 +48,11 @@ export class TraitDAO {
   }
 
   static async updateTraits(programId: string, newTraits: Trait[]): Promise<BiResponse> {
-    /*const { data } =  await api.call({
+    const { data } =  await api.call({
       url: `${process.env.VUE_APP_BI_API_V1_PATH}/programs/${programId}/traits`,
       method: 'put',
       data: newTraits
-    }) as Response;*/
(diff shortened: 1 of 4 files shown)


nickpalladino (commented):
Was playing around editing traits and when I switched from text to ordinal I get a 422 error. Looks like bi-web is not sending categories in the scale info.

image

timparsons (approved):
Approving contingent on addressing the comment Nick made earlier about the `422`.

ctucker3:
> Was playing around editing traits and when I switched from text to ordinal I get a 422 error. Looks like bi-web is not sending categories in the scale info.
> 
> image

Good catch! Vue wasn't catching the array update.

nickpalladino (commented):
The text -> ordinal thing works but I'm seeing some other things on the edit form, maybe we just create a new card to fix bugs with the form? If I try to edit and change from ordinal to nominal, it doesn't seem to change.

ctucker3:
> The text -> ordinal thing works but I'm seeing some other things on the edit form, maybe we just create a new card to fix bugs with the form? If I try to edit and change from ordinal to nominal, it doesn't seem to change.

Yeah I like that idea. Made a card fir it:

https://breedinginsight.atlassian.net/browse/BI-735

nickpalladino (approved):

PR #76, opened 2021-02-08
[BI-698] Import Traits not available until refreshing page after adding self to program

dmeidlin:
If the active user is being added to a program then the active user roles are updated as part of the success handler.

Description: Import Traits not available until refreshing page after adding self to program
Component/Version: v0.2
Steps to Reproduce

login as admin

add self to a program

select traits in left navigation bar

Expected result:
Import Traits visible as a selection 

Actual Result:
Import traits not visible until the page is refreshed

### DIFF
--- a/src/components/program/ProgramUsersTable.vue
+++ b/src/components/program/ProgramUsersTable.vue
@@ -192,6 +192,9 @@
   import { User } from '@/breeding-insight/model/User';
   import {UserService} from "@/breeding-insight/service/UserService";
   import { DataFormEventBusHandler } from '@/components/forms/DataFormEventBusHandler';
+  import store from "@/store";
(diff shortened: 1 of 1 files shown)


timparsons (approved):
Looks good.  Approving contingent on fixing the build error.

ctucker3 (approved):

DebWeigand (changes requested):
When I test this, following the exact steps to reproduce, I can pass this.  
But, if I add myself as a 'member' first, and then later edit my role to 'breeder', then I still have to refresh the page to see 'Import Traits'.

dmeidlin:
Ok, the active user role should be updated now when the active user role is edited in program management.

DebWeigand (approved):

PR #78, opened 2021-02-19
Feature/bi 564

dmeidlin:
### Archived Traits Table

Hey! Please take a look at how Vuex can be used and make any comments. The purpose of this PR is to make it easy to make comments and discuss specifics of how vuex could be used for biweb. This is not intended to be a final review before merging.

#### Required State Data
All traits in a program are pulled from the backend and stored locally. Filtering the existing all-traits collection for archived traits will not suffice because all-traits data stored is only one page of the collection, and the archived traits can be distributed over all pages.
An additional GET archived-traits endpoint could eliminate the need to store an entire collection on the frontend, but that doesn't exist, now.

The pagination and pagination controls state for the archive table must be independent of that of all other tables.

#### Vuex
Vuex modules for traits, archive, and pagination were added to the store with the hierarchy: 
```
root --> traits --> archive --> pagination
             \
              \--> pagination
```
The all-traits collection is stored in the `traits` module; the archived-traits and paginated-archived-traits are stored in the `archive` module; and the pagination state and control data is stored in the respective `pagination` module, although only the archive controls are currently set up.

You can explore the state tree using the state tab on the chrome debugging tool Vue extension.  Also, you will see a history of mutations committed as well as a list of getters and mutations in each module.

#### Vuex-specific Component Changes
`@/components/trait/TraitsArchiveTable.vue` was necessarily created for the new table but also the SidePanelTable and child components were copied over into `@/components/tables/ArchiveSidePanelTable.vue`, `@/components/tables/ArchiveBaseTable.vue`, and `@/components/tables/ArchivePaginationControls.vue`. These last three files are temporary to hold the changes made to these child components in order to use Vuex for state management for the archive table.

The props and methods originally used for the all-traits table was ported into vuex module state, getters, mutations, and actions which are then added to an archive component using the vuex helper functions `mapState`, `mapGetters`, `mapMutations`, and `mapActions`.  State and getters are mapped to component computed values; mutations and actions are mapped to component methods.

For input forms like the page size selector box two-way data binding is not allowed since the store operates in strict mode. `v-model` is simply replaced with `v-bind` and the existing `v-on:change`.

The pagination event listeners don't need to bubble up from the ArchivePaginationControls to the TraitsArchiveTable component level so those have been removed from the top level.

#### Work in Progress
This branch is only supposed to illustrate how Vuex could be used to manage state on the front end as the UI gains in complexity in order to make any discussions and decision specific and meaningful. I only went as far as getting the pagination controls working, but Vuex would also need to manage the all-traits data and pagination as well as sidebar panel state information for the table to function property. For example, editing traits at the all-traits table will not affect the archived traits.

Also, the archived traits right now only return all traits since the filtering has been commented out since the new backend features supporting the active prop haven't been merged in yet.

### DIFF
--- a/src/components/tables/ArchivePaginationControls.vue
+++ b/src/components/tables/ArchivePaginationControls.vue
@@ -1,0 +1,111 @@
+<template>
+  <b-pagination
+      v-if="pagination"
+      :total="pagination.totalCount"
+      :current="pagination.currentPage"
+      range-before="1"
+      range-after="1"
+      order="is-centered"
+      size="is-small"
+      :simple="false"
+      :rounded="false"
+      :per-page="pagination.pageSize"
+      aria-next-label="Next page"
+      aria-previous-label="Previous page"
+      aria-page-label="Page"
+      aria-current-label="Current page"
+      v-on:change="updatePage($event)"
+  >
+    <b-pagination-button
+        slot="previous"
+        slot-scope="props"
+        :page="props.page"
+        tag="a"
+    >
+      Previous
+    </b-pagination-button>
+
+    <template
+        slot="next"
+        slot-scope="props"
+    >
+      <b-pagination-button
+          :page="props.page"
+          tag="a"
+      >
+        Next
+      </b-pagination-button>
+
+      <div class="pagination-extras">
+        <div class="page-size-select pagination-link">
+          <div class="select is-small">
+            <label for="paginationSelect" class="is-sr-only">Results Per Page</label>
+            <select
+                id="paginationSelect"
+                :value="pagination.pageSize"
+                v-on:change="updatePageSize($event.target.value)"
+            >
+              <option value="10">
+                10
+              </option>
+              <option value="20">
+                20
+              </option>
+              <option value="50">
+                50
+              </option>
+              <option value="100">
+                100
+              </option>
+              <option value="200">
+                200
+              </option>
+            </select>
+          </div>
+          <span>per page</span>
+        </div>
+
+        <a
+            data-testid="showAll"
+            role="button"
+            class="pagination-link show-all-button"
+            v-bind:class="{ 'has-background-info': pagination.totalPages === 1}"
+            v-on:click="toggleShowAll"
+        >
+          Show All
+        </a>
+      </div>
+    </template>
+  </b-pagination>
+</template>
+
+<script lang="ts">
+  import {Component, Prop, Vue} from 'vue-property-decorator'
+  import {Pagination} from "@/breeding-insight/model/BiResponse";
+  import {TOGGLE_SHOW_ALL, UPDATE_PAGE, UPDATE_PAGE_SIZE} from "@/store/pagination/mutation-types";
+  import {mapMutations} from 'vuex';
+
+  @Component({
+    methods: {
+      ...mapMutations({
+        [UPDATE_PAGE_SIZE](commit, newSize) {
+          return commit(this.namespace + UPDATE_PAGE_SIZE, newSize);
+        },
+        [UPDATE_PAGE](commit, newPage) {
+          return commit(this.namespace + UPDATE_PAGE, newPage);
+        },
+        [TOGGLE_SHOW_ALL](commit) {
+          return commit(this.namespace + TOGGLE_SHOW_ALL);
+        }
+      })
+    }
+  })
+  export default class PaginationControls extends Vue {
+
+    @Prop()
+    private pagination!: Pagination;
+    @Prop()
+    private namespace?: string;
+
+  }
+</script>
--- a/src/components/trait/TraitsArchiveTable.vue
+++ b/src/components/trait/TraitsArchiveTable.vue
@@ -149,7 +149,7 @@ import WarningModal from '@/components/modals/WarningModal.vue'
 import {PlusCircleIcon} from 'vue-feather-icons'
 import {validationMixin} from 'vuelidate';
 import {Trait} from '@/breeding-insight/model/Trait'
-import {mapMutations, mapGetters, mapState, mapActions} from 'vuex'
+import {mapGetters, mapState, mapActions} from 'vuex'
 import {Program} from "@/breeding-insight/model/Program";
 import NewDataForm from '@/components/forms/NewDataForm.vue'
 import BasicInputField from "@/components/forms/BasicInputField.vue";
@@ -158,7 +158,7 @@ import TraitDetailPanel from "@/components/trait/TraitDetailPanel.vue";
 import {TraitService} from "@/breeding-insight/service/TraitService";
 import EmptyTableMessage from "@/components/tables/EmtpyTableMessage.vue";
 import TableColumn from "@/components/tables/TableColumn.vue";
-import {Metadata, Pagination} from "@/breeding-insight/model/BiResponse";
+import {Metadata} from "@/breeding-insight/model/BiResponse";
 import { StringFormatters } from '@/breeding-insight/utils/StringFormatters';
 import { TraitStringFormatters } from '@/breeding-insight/utils/TraitStringFormatters';
 import BaseTraitForm from "@/components/trait/forms/BaseTraitForm.vue";
@@ -168,7 +168,7 @@ import {MethodClass} from "@/breeding-insight/model/Method";
 import {DataType, Scale} from "@/breeding-insight/model/Scale";
 import {SidePanelTableEventBusHandler} from "@/components/tables/SidePanelTableEventBus";
 import { DataFormEventBusHandler } from '@/components/forms/DataFormEventBusHandler';
-import {TOGGLE_SHOW_ALL, UPDATE_PAGE, UPDATE_PAGE_SIZE} from "@/store/pagination/mutation-types";
+import {GET_ALL_ARCHIVED_TRAITS, GET_ARCHIVED_TRAITS} from "@/store/traits/archive/action-types";
 
   @Component({
   mixins: [validationMixin],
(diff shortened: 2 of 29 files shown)


timparsons:
@dmeidlin This is really helpful to see a working example in our codebase of Vuex, thank you!!  When a user navigates away from the archive table, then back to it later on, does the state get reset? Is the reset a responsibility of the Vuex module or of the component utilizing the Vuex module?

I see two different types of state that we currently have -> 1) system state like the user's auth info, and 2) page state, information needed for a specific page to load.  For the second type, what advantages does Vuex have over the state being held in the page component?

nickpalladino:
> An additional GET archived-traits endpoint could eliminate the need to store an entire collection on the frontend, but that doesn't exist, now.

I think you should be able to use `/programs/{programId}/traits/search` to get the archived traits.

nickpalladino:
> I see two different types of state that we currently have -> 1) system state like the user's auth info, and 2) page state, information needed for a specific page to load. For the second type, what advantages does Vuex have over the state being held in the page component?

Some additional thoughts, if we do the right side panel we could have whatever is loaded in that persist independently of the loaded page content potentially.

src/components/trait/TraitsArchiveTable.vue
```diff
@@ -0,0 +1,362 @@
...
+  private traitSidePanelState: SidePanelTableEventBusHandler = new SidePanelTableEventBusHandler();
+
+  // Edit form
+  private editValidationHandler: ValidationError = new ValidationError();
+
+  // Archive trait
+  private deactivateActive: boolean = false;
+
+  mounted() {
+    this[GET_ARCHIVED_TRAITS]();
```
dmeidlin, line 226: > @dmeidlin This is really helpful to see a working example in our codebase of Vuex, thank you!! When a user navigates away from the archive table, then back to it later on, does the state get reset? Is the reset a responsibility of the Vuex module or of the component utilizing the Vuex module?
> 
No, state is not reset.
When coming back to the table the `getArchivedTraits` action is dispatched via the mounted hook.  If the the archivedTraits collection is empty or a refresh flag is sent as a payload(currently not), then the archived traits are pulled from the allTraits collection.

The component using the vuex module is responsible for a reset.
Any mutations to the state, like a reset, would have to be committed in the component or committed as part of an action dispatched by a component.  Mutations and actions are mapped to the component as methods.

dmeidlin:
> I see two different types of state that we currently have -> 1) system state like the user's auth info, and 2) page state, information needed for a specific page to load. For the second type, what advantages does Vuex have over the state being held in the page component?

The questions is how to handle communication in bi-web between components that must share access to component state data. This PR involves one type of scenario that already occurs in the app; the pagination controls component is deeply nested in the tree below the archive table component. In general, we will have the need also arise between sibling components or between grandchildren components on different roots as well. So we are talking about what to do when:
multiple sibling components use the same data; or
multiple roots need data access; or
there is deep nesting of components.
The goal is to avoid the clutter of having to pipe state data through intermediary components.

We are currently using Vue version 2.6.10.  In version 2 there were only two real ways to handle the component state access: an event bus or vuex.  Vue 3 has deprecated support for event busses:https://v3.vuejs.org/guide/migration/events-api.html#_3-x-update. So, going forward with event buses would require another approach after upgrading to the next major version of Vue.

Alternatively, Vue 3 enables inter-component access to state via `Provide/Inject`. This is fine for small apps but doesn't scale well since tracing where data was provided can be challenging with multiple devs contributing. It would also require us to upgrade to Vue 3.

Vue 3.0 was released in Sept. 2020.  All major versions of Vue support Vuex, so by using Vuex there is no need to change how we handle state management and inter-component communication when we eventually upgrade to the next major version. Also, there is no need to upgrade now in order to use alternatives that are only available in the latest major version.

In this example of the archive table with paginations controls using Vuex to manage state the pagination methods and event handlers are contained in the pagination component and module and the the archiveTable component and module contain the archive logic and data, with the namespace being passes as a prop to the pagination component and watches setup to describe how to respond to pagination control events. Intermediary components like the sidepanel table don't need to re-emit pagination events up to the archive table. So, it generally reduces the need to pipe through intermediaries and aids in encapsulating components better.

Mapping the state data and getters to computed values allows updates using Vue's reactive flow, which can make updates to components easier.

Also, the chrome vue extension makes it super easy to explore the state, view a history of mutations, list any getters in addition to the component tree and history of events fired. You can also reset the state to what it was when a mutation occurred, so it is really helpful for debugging and exploring new features. I'm sure you could do most of the same with IDE debugging features; it just is a convenience to have the plugin. But I think that also reflects the general high level of continued support for Vuex.

If component state data does not need to be shared then Vuex wouldn't be necessary, but the use of observables would be very useful.
Observables are available starting with Vue 2.6, and should be used for managing state needed within a single component:
https://vuejs.org/v2/api/#Vue-observable

src/components/tables/ArchivePaginationControls.vue
```diff
@@ -0,0 +1,111 @@
...
+      ...mapMutations({
+        [UPDATE_PAGE_SIZE](commit, newSize) {
+          return commit(this.namespace + UPDATE_PAGE_SIZE, newSize);
+        },
+        [UPDATE_PAGE](commit, newPage) {
+          return commit(this.namespace + UPDATE_PAGE, newPage);
+        },
+        [TOGGLE_SHOW_ALL](commit) {
+          return commit(this.namespace + TOGGLE_SHOW_ALL);
+        }
```
ctucker3, line 99: I like the idea of these event driven methods being laid out at the top of the component so we can see what state methods the page uses. This is in contrast to the event bus which has the state event calls scattered among the methods.

src/components/trait/TraitsArchiveTable.vue
```diff
@@ -0,0 +1,362 @@
...
+  computed: {
+    ...mapGetters([
+      'activeProgram'
+    ]),
+    ...mapGetters('traits/archive',[
+        'archivePagination'
+    ]),
+    ...mapState('traits/archive', {
+      paginatedArchivedTraits: state => state.paginatedTraits,
+      paginationController: state => state.pagination
```
ctucker3, line 188: If you had two tables on a page, would you be able to have separate pagination controls for both of them?

src/store/pagination/index.ts
```diff
@@ -0,0 +1,38 @@
...
+import {mutations} from '@/store/pagination/mutations';
+import {PaginationControllerState} from '@/store/pagination/types';
+import {RootState} from '@/store/types';
+
+export const state: PaginationControllerState = {
+    currentPage: 1,
+    pageSize: 50,
+    showAll: false,
+    currentCall: undefined
+};
```
ctucker3, line 29: Say you are on the traits page and select a `page size` of `100`. If you go to the programs page, will the page size be reset to `50` or will it be `100` still?

ctucker3 (commented):
Just finished reviewing, so I am going to record some of my thoughts here so I can remember them for the meeting tomorrow. 

For an analysis of this vs the state bus, here are the major thoughts I had, broken down by category. These are presented through the filter of my perception, so things I present as facts may only be facts in my world!

1. Ease/Size/Speed of Implementation
 - For me the state bus wins in this category, it is one class that contains all of the state variables and methods in that class. Vuex is more formalized so it contains more code/files to get things running. For example, for the archive traits state, there was 7 files to get that set up. 

2. Code style/Formalization
- Vuex wins in this category. I liked the formal definition of state events in the top of the component, it made it clear what the state methods were and which ones the component used. For the state bus, the methods are called on the state bus class and are nested within the component methods.

3. Complexity/Probability of Mistakes
- Because the state bus is not as formalized, it is simpler to implement because you can be a little more free with your implementation. There is a rough pattern, but anything goes, so we are responsible for making sure the code makes sense. This makes the implementation simpler, but probability of mistakes is completely dependent on our implementation and reviewing skills. Being more formalized, Vuex is likely less prone to error. But, since everything is in separate files, it will probably be harder for reviewers to trace the logic. I know it took some effort on my part during this review, but it was also my first time looking at this pattern so it was mostly new to me. 

Potential Issues
- This category is just for the Vuex state management since I've already worked with the state bus. Since Vuex shares a state between all of the components in the Vue session, we need to be more careful about resetting state on component loads. It also seems like we would run into issues if we wanted to have two tables of the same kind on the same page. This might seem rare, but I'm wondering if we will run into issues like this when we make reporting pages. Where maybe we have two trait tables display with different filters applied on them, or something along those lines.

src/components/tables/ArchivePaginationControls.vue
```diff
@@ -0,0 +1,111 @@
...
+      </b-pagination-button>
+
+      <div class="pagination-extras">
+        <div class="page-size-select pagination-link">
+          <div class="select is-small">
+            <label for="paginationSelect" class="is-sr-only">Results Per Page</label>
+            <select
+                id="paginationSelect"
+                :value="pagination.pageSize"
+                v-on:change="updatePageSize($event.target.value)"
```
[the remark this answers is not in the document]
dmeidlin: I agree, and I gave up trying after making several attempts to use the UPDATE_PAGE_SIZE string variable instead of the camelCase variable name. It was really annoying! I tried [UPDATE_PAGE_SIZE](...) and this[UPDATE_PAGE_SIZE](...) and similar variations, but nothing worked to call the method. There is something with how vue is interpreting the JS between quotes that I haven't figured out, yet.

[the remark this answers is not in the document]
dmeidlin: FYI, Gitub formatted away the square brackets that wrap around the UPDATE_PAGE_SIZE string in what I wrote above.

src/components/trait/TraitsArchiveTable.vue
dmeidlin: Yes, the state data is namespaced. In this case the namespace used is 'traits/archive'. So this pagination control state is not accessible globally like program and user data. If there is another pagination control for another table, you would call teh same mapState() helper function with the namespace 'traits/favorite', for example. In addition, the prop names allow you to rename the methods for clarity, so even though the pagination module is being reused you can have different names in the component like archivePaginationController and favoritePaginationController each mapped to their respective state.pagination.

src/components/tables/ArchivePaginationControls.vue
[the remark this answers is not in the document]
dmeidlin: `[UPDATE_PAGE_SIZE](...)` and `this[UPDATE_PAGE_SIZE](...)`

src/store/pagination/index.ts
dmeidlin: Using Vuex, pagination states for different tables (traits and programs) are independent of each other, so if the program table had page size = 50 before traits table page size was set to 100 then programs would still have page size = 50.

PR #79, opened 2021-02-24
BI-690 - Edit existing trait which has observation data

nickpalladino:
Possible future improvements (don't want to spend more time on this currently):
- caching editable status
- cancelling editable requests when table row is changed

### DIFF
--- a/src/components/trait/TraitListsTable.vue
+++ b/src/components/trait/TraitListsTable.vue
@@ -403,7 +403,7 @@ export default class TraitTable extends Vue {
         const deletions: string[] = this.processValidationErrors(this.editValidationHandler, this.editTrait!);
         this.$emit('show-error-notification', `Error updating trait. ${this.editValidationHandler.condenseErrorsSingleRow(deletions)}`);
       } else {
-        this.$emit('show-error-notification', 'Error updating trait.');
+        this.$emit('show-error-notification', error);
       }
     } finally {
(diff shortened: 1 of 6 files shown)


timparsons (approved):

src/components/trait/TraitListsTable.vue
```diff
@@ -268,6 +275,22 @@ export default class TraitTable extends Vue {
...
+  async editable(trait: Trait) {
+    let traitEditable = false;
+    this.loadingTraitEditable = true;
+    try {
+      const [editable] = await TraitService.getTraitEditable(this.activeProgram!.id!, trait.id!) as [boolean, Metadata]
+      traitEditable = editable;
+      this.currentTraitEditable = traitEditable;
+    } catch (error) {
+      // Display error that traits cannot be loaded
+      this.$emit('show-error-notification', 'Error while trying to load traits');
```
ctucker3, line 287: This error message seems incorrect. This is for if the editable endpoint fails right?

ctucker3 (approved):
One small comment, so approving since I am good with this PR once that is fixed.

src/components/trait/TraitListsTable.vue
nickpalladino: ya, thanks I'll update that

nickpalladino: pushed changes

PR #80, opened 2021-02-24
[BI-745] - Prototype BrAPI importer

ctucker3:
EDIT: This should be run with the backend now. The endpoint to get all import configs is complete so the mock for that was removed. 

A prototype of the brapi importer. It runs on mock data right now, but contains the skeleton to connect to the backend when the backend is implemented. 

For review, feel free to leave comments on design recommendations. Keep in mind that it is not meant to have a polished look, but is meant to look good enough. We will likely have the coordinators use this version for testing, so if something looks real wonky, we might want to fix that. 

Below are some improvements I would like to make in the future. I will make a card for these when we are ready to merge this card in:

- Make UI pretty
    - Spacing, alignment, colors, notifications
- Have summary one liner to put in the title of the cards (ex. show file name on import step card once complete)
- Add descriptions for each step
- Improve language to make the process more clear to the user
- Add a mapping summary on mapped objects (ex. show attribute value and attribute name in summary of cross attribute)
- Make mapping fields prettier (the relationship card needs some work. Others too probably).
- Prototype a real-time lookup relationship (example look for a trait in the db when mapping an observations file)
- Save to local storage, so if they hit the back button, it doesn't lose all of their data


nickpalladino:
image

When editing the mappings, maybe there could be a button for saving? I wasn't sure if I navigated back using the link at the top if I would lose my settings.

dmeidlin:
Nice job! It looks like a great start. I like the bread crumbs at the top. Do you have plans to do more with the Reminders card? It seems a little redundant as it is since the drop down menus already list all of the column headings. Displaying the headings earlier in the workflow might make more sense, like in the summary of the "step 2: upload file" card. The idea being that it's a check before proceeding to step 3. I like the use of "show info" buttons, too.

src/views/import/BrAPIImporter.vue
```diff
@@ -0,0 +1,610 @@
...
+    private mapping: ImportMappingConfig = new ImportMappingConfig(undefined);
+    private importMappings: ImportMappingConfig[] = [];
+    private importMappingOptions: any[] = [];
+
+    private currentStep = 1;
+    private state = ImportState.CHOOSE_IMPORT;
+    private importStateMachine = createMachine({
+        id: 'import',
+        initial: ImportState.CHOOSE_IMPORT,
+        states: {
```
dmeidlin, line 368: After taking a look at https://github.com/Breeding-Insight/bi-web/pull/78 and how Vuex was used there we should discuss our options for managing access to state data like this. The BrapiImporter branch of the component tree will only continue to grow, so it would be nice to discuss this soon since whatever we ultimately decide is the best practice, will be consistent going forward.

ctucker3:
> Nice job! It looks like a great start. I like the bread crumbs at the top. Do you have plans to do more with the Reminders card? It seems a little redundant as it is since the drop down menus already list all of the column headings. Displaying the headings earlier in the workflow might make more sense, like in the summary of the "step 2: upload file" card. The idea being that it's a check before proceeding to step 3. I like the use of "show info" buttons, too.

Yeah I agree that the reminder is a bit redundant right now. I was hoping to somehow show which columns had been used so far. Like a green check in the tag for that column, or something along those lines. Since the mapping will take a while for the coordinators to set up, it would be nice to show them where they are in the mapping so far in all of the steps. It doesn't necessarily have to be with the reminder card, but that was a first whack at it. 

> When editing the mappings, maybe there could be a button for saving? I wasn't sure if I navigated back using the link at the top if I would lose my settings.

Yeah I think that is a good move. I think that will improve the flow if when they click the button is takes them back to the parent object page.

src/views/import/BrAPIImporter.vue
ctucker3: Yep, open to discussing options! This state machine is only used in this view component, so I would think this could just stay in here as it is. But, for data being passed to and from the children components that could use either the state bus or vuex.

PR #87, opened 2021-04-20
[BI-741] - Adding BrAPI Info Page

timparsons:
Adding page to display BrAPI connection information, as well as a QR code for use in Field Book to connect to BI


nickpalladino (approved):
I tried the Field Book scan a server barcode feature. It did populate the BrAPI Base URL. It also tried to load the site in chrome which failed for my dev server since it's a localhost url. I'm guessing it was trying to authorize so that's good. One thing with the Breeding Insight BrAPI page, I remember Liz saying users shouldn't be seeing BrAPI really. Maybe we can make it a bit more user friendly in the future. Could be called third party app integration or something and then maybe an explanation of what BrAPI is.

dmeidlin (approved):
This displayed the correct brapi base url for the current program and the QR-code displayed scanned correctly. Nice!

PR #92, opened 2021-05-28
[BI-846] - Hiding create/edit buttons for traits for member role

timparsons:
Added abilities of `create` and `update` of `Trait` to the `breeder` role, and then enforced them by hiding or showing create, edit, and archive buttons (breeder can see the buttons, member cannot).

### DIFF
--- a/src/components/trait/TraitDetailPanel.vue
+++ b/src/components/trait/TraitDetailPanel.vue
@@ -69,11 +69,11 @@
       <p class="has-text-weight-bold mt-3 mb-0">Description of collection method</p>
       <p>{{data.method.description}}</p>
 
-      <ProgressBar v-if="loadingEditable" v-bind:label="'Checking trait editability status'"
+      <ProgressBar v-if="loadingEditable && $ability.can('update', 'Trait')" v-bind:label="'Checking trait editability status'"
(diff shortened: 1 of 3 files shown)


nickpalladino (commented):
Did you test to see if you change the program user role that the create/edit work without refreshing. I know we've had stuff like that happen in other areas.

timparsons:
> Did you test to see if you change the program user role that the create/edit work without refreshing. I know we've had stuff like that happen in other areas.

I just tested this, and it does require a hard refresh or the user logging out and then back in to take affect.  I was logged in on one browser as a system admin, and on another browser as just a member of a program.  As the system admin, I modified the other user to be in the breeder role, then switched to the other browser and had to do a hard refresh for the role change to take affect.  I think fixing that is outside the scope of this card, and I'm not sure if it's something we want to change (lots of things require you to log out then back in after a permission change).

nickpalladino:
Ya, I guess for a typical user it won't matter since they can't change the role themselves. As a sys admin i know we had the one where you add yourself to a program and didn't see the import traits menu option unless you refreshed. I guess it'd be a similar thing mostly applicable to sys admin users. @DebWeigand

nickpalladino (approved):

dmeidlin (approved):

PR #97, opened 2021-07-06
[BI-972] - Testing Automation Framework (TAF) Button ID selectors

HMS17:
Adding IDs to buttons to make TAF testing more robust (as opposed to using xpaths for selectors).

- Download the Trait Import Template
- Choose a file…
- Choose a different file…
- Import
- Yes, abort
- Log out
- Home
- Import File
- Program Management
- Traits
- BrAPI
- Users
- Programs

Convention: lowercase, separated by dashes, prefaced with component name, ie: “fileselector-choose-file”

Linked to https://github.com/Breeding-Insight/taf/pull/3 which modifies associated TAF tests

### DIFF
--- a/src/components/file-import/FileSelectMessageBox.vue
+++ b/src/components/file-import/FileSelectMessageBox.vue
@@ -79,7 +79,7 @@
           <div class="level-right">
             <div class="level-item">
               <div>
-                <a v-if="file" class="button is-primary has-text-weight-bold" v-on:click="$emit('import')">Import</a>
+                <a v-if="file" class="button is-primary has-text-weight-bold" :id="importButtonId" v-on:click="$emit('import')">Import</a>
               </div>
             </div>
           </div>
@@ -111,6 +111,8 @@
 
     @Prop()
     private errors!: ValidationError | string | null;
+
+    private importButtonId: string = "import-button";
 
     mounted() {
       this.file = this.value;
--- a/src/components/file-import/FileSelectMessageBox.vue
+++ b/src/components/file-import/FileSelectMessageBox.vue
@@ -112,7 +112,7 @@
     @Prop()
     private errors!: ValidationError | string | null;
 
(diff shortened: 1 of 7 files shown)


src/components/file-import/FileSelectMessageBox.vue
```diff
@@ -112,6 +112,8 @@
     @Prop()
     private errors!: ValidationError | string | null;
 
+    private importButtonId: string = "import-button";
```
nickpalladino, line 115: Do we need to worry about collisions of id names across components? Should there be a naming convention including the component name?

HMS17: It's something I was wondering about since I'm a little less familiar with BI. I'm open to establishing a naming convention (beyond going with lowercase and dashes for id names).

nickpalladino: ya, not sure what the best approach is. In some of our old unit tests we were using the vue test utils and had a data-testid. @ctucker3 did most of the work on that and may have some input.

HMS17: Yeah in my research it seemed there were mixed opinions on using id vs some custom attribute.

ctucker3: I don't have much of a preference either way. If we go with id, it makes sense to me to include the component name in some form in the id for the reusable components.

timparsons: Given that @HMS17 has done a bunch of work using IDs, I'd also vote to include component name in with the id value.

HMS17: I added component name to the ids I have so far.

nickpalladino (approved):

timparsons (approved):

PR #100, opened 2021-07-09
[BI-895] - Convert the program user's table to use Buefy data tables

timparsons:
The program user's table now uses Buefy's Table component (https://buefy.org/documentation/table/).  Rather than implementing Buefy's table directly in `ProgramUsersTable` or modifying `ExpandableRowTable`, a new component named `ExpandableTable` was created.  The main reason for not modifying `ExpandableRowTable` was to limit the scope of impact to just `ProgramUsersTable` rather than the entire UI, and to allow for each table currently using that component to be migrated and tested individually and incrementally.

The `ProgramUsersTable` should have feature parity with the original implementation.  In addition, columns can now be sorted.

All features of Buefy Table should be accessible to client code using `ExpandableTable`, as the `$attrs` are passed through to Buefy Table.  As an example, Buefy Table allows configuration to indicate if data is loading, but `ExpandableTable` does not have to have an explicit property defined to set this (`ProgramUsersTable` does set the `loading` value).

*Known bugs*
- if there are 11 records, and you change the page size from 50 to 10, when navigating to the second page, the first record is highlighted as a new record.  I tested this on the ProgramLocationsTable, and this same bug appears there (which is using `ExpandableRowTable`)
- If records are paginated, sorting only works on the current page

*Future Development*
- Convert all existing `ExpandableRowTable` usages to `ExpandableTable`
- Create a component for the side panel table that uses Buefy table


nickpalladino (commented):
Do we have cards for the identified bugs and future work?

timparsons:
> Do we have cards for the identified bugs and future work?

I've created issues for the known bugs, and will create stories for future development next week.  Thanks!

nickpalladino (approved):

dmeidlin (approved):

PR #103, opened 2021-07-28
[BI-1030] - Added classes for modal text

HMS17:


src/components/admin/AdminProgramsTable.vue
```diff
@@ -23,7 +23,7 @@
       v-on:deactivate="deactivateActive = false"
     >
       <section>
-        <p class="has-text-dark">
+        <p class="has-text-dark" :class="modalTextClass">
```
ctucker3, line 26: Not a big deal for this PR, but we are using explicit calls for binding and the such, ex. `v-bind:modalTextClass`. Works the same, its just the pattern we started following a while back.

src/components/admin/AdminProgramsTable.vue
```diff
@@ -243,6 +243,8 @@ export default class AdminProgramsTable extends Vue {
 
   private serverError: FieldError[] = [];
 
+  private modalTextClass: string = "modal-text";
```
ctucker3, line 246: Can this be put in the main vue class, or in another class so that all of the components can use the same variable instead of having to define it in each?

src/components/modals/WarningModal.vue
```diff
@@ -55,6 +55,8 @@ export default class WarningModal extends Vue {
   active!: boolean;
   @Prop()
   msgTitle! : string;
+
+  private modalHeaderClass: string = "modal-header";
```
ctucker3, line 59: See comment above about common class for these name constants.

ctucker3 (changes requested):

src/components/admin/AdminProgramsTable.vue
HMS17: I thought we were using the other one as the standard, since that is what I was using for my other class and ID JIRAs. Should I make a card for changing all my prior changes or just keep it in mind moving forward?

src/components/admin/AdminProgramsTable.vue
HMS17: main class as in App.vue or something else?

src/components/modals/WarningModal.vue
HMS17: I think it might be okay here since it's only being defined in one place? could admittedly also move it.

src/components/admin/AdminProgramsTable.vue
ctucker3: I would just keep it in mind going forward since it doesn't effect anything functionally.

src/components/modals/WarningModal.vue
ctucker3: Yep, agreed. Probably fine for this one.

src/components/admin/AdminProgramsTable.vue
HMS17: Placed class name in main.ts

ctucker3 (approved):

dmeidlin (approved):

PR #104, opened 2021-08-02
[BI-1039] Switch label and value in scale categories

ctucker3:
Changes the scale categories README description to say `<value of the category> = <label of the category>` instead of `<label of the category> = <meaning of the category>`.

Changes to the UI:
- Switch the label and value around. The UI will look the same, values are just assigned differently to the scale model.
- Changes for switching between nominal and ordinal:
  - When switching from ordinal to nominal, the category value is set equal to the previous category label. 
  - When switching from nominal to ordinal, the category label is set equal to the category value.
    - These changes were made to mimic the previous behavior on the UI. 
  - Cleared the server validations for categories in the front end when switching between nominal and ordinal
    - The validations were confusing because the errors seemed to move fields because of the orientation of the fields. 

Changes to BreedBase:
- Changes brapi v1 categories to be formatted as "value=label" instead of "label=value" to follow the switching on our side. 
- If a label isn't passed, the label is set as the value, instead of being set as an incrementing label. 
  - This fixes nominal categories automatically generating labels.

Changes to bi-api:
- There are no changes to the backend

Changes to Field Book:
- Field Book now uses the `value` in the brapi category, instead of the `label`

This PR is related to some FieldBook work being done, https://github.com/PhenoApps/Field-Book/issues/294.
Here is the Field Book PR: https://github.com/PhenoApps/Field-Book/pull/320

Chaney is going to do most of that work, although I have a PR for a quick fix to match the brapi v1 implementation in FieldBook.

Things you'll have to do:
- Update your trait files to be label = value
- Run the migration script below to fix your brapi v2 categories in breedbase
- Fix your nominal traits if you have any and if you want to.

You will have to migrate any existing categorical traits in BreedBase because `label` and `value` are switched currently. Here is a query to do that:

```
DO
$do$
declare
    category_row record;
begin
	for category_row in 
		select label.cvtermprop_id as label_id, label.value as label_value, value.cvtermprop_id as value_id, value.value as value_value 
		from cvterm 
		join cvtermprop as label on cvterm.cvterm_id = label.cvterm_id and label.type_id = 79072
		join cvtermprop as value on cvterm.cvterm_id = value.cvterm_id and label.rank = value.rank and value.type_id = 79073
		where 
		cv_id = 79
	loop
		RAISE NOTICE 'label_id: %, label: %, value_id: %, value: %', category_row.label_id, category_row.label_value, category_row.value_id, category_row.value_value;
		update cvtermprop set value = category_row.value_value where cvtermprop.cvtermprop_id = category_row.label_id;
		update cvtermprop set value = category_row.label_value where cvtermprop.cvtermprop_id = category_row.value_id;
		--select category_row.value_id, category_row.value_value from category_row;
	end loop;
end;
$do$;
```

### DIFF
--- a/src/components/trait/TraitImportTemplateMessageBox.vue
+++ b/src/components/trait/TraitImportTemplateMessageBox.vue
@@ -32,7 +32,7 @@
             <div class="level-item">
               <div class="has-text-dark has-text-centered is-size-7">
                 <!-- temporary link until the backend card is done -->
-                <a href="https://cornell.box.com/shared/static/g6vmpktevwolwkyatwxbdbzpkbwv77ov.xls"
+                <a href="https://cornell.box.com/shared/static/8sp0qvccpjotosiv8576tczeg09nnvao.xls"
                   class="button is-outlined is-primary" :id="downloadTraitTemplateId">Download the Trait Import Template</a>
                 <br/>Template version placeholder
               </div>
--- a/src/breeding-insight/model/errors/ValidationError.ts
+++ b/src/breeding-insight/model/errors/ValidationError.ts
@@ -75,4 +75,14 @@ export class ValidationError {
     }
     return errorSentence;
   }
+
+  clearValidation(rowIndex: number, errorName: string) {
+    if (this.rows && this.rows[rowIndex]) {
+      for (const [errorIndex, field] of this.rows[rowIndex].errors!.entries()) {
+        if (field.field === errorName) {
+          delete this.rows[rowIndex].errors![errorIndex];
+        }
+      }
+    }
+  }
 }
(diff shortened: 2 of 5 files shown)


timparsons (approved):

src/components/trait/TraitImportTemplateMessageBox.vue
```diff
@@ -32,7 +32,7 @@
             <div class="level-item">
               <div class="has-text-dark has-text-centered is-size-7">
                 <!-- temporary link until the backend card is done -->
-                <a href="https://cornell.box.com/shared/static/g6vmpktevwolwkyatwxbdbzpkbwv77ov.xls"
+                <a href="https://cornell.box.com/shared/static/8sp0qvccpjotosiv8576tczeg09nnvao.xls"
```
nickpalladino, line 35: looks like there's still a meaning of category for nominal, is this what you intended or did you want to use the new terminology?

nickpalladino (commented):
Should that query maybe be moved to the breedbase_configs repo or some centralized place so we can keep track of all the data migrations we need to do?

src/components/trait/TraitImportTemplateMessageBox.vue
ctucker3: Good catch, I updated the file.

nickpalladino:
> Should that query maybe be moved to the breedbase_configs repo or some centralized place so we can keep track of all the data migrations we need to do?

One this is addressed I'll approve

nickpalladino (approved):

PR #106, opened 2021-08-16
 [BI-982] - Ontology BrAPI Mapping for Trait

dmeidlin:
There is a new trait import template required for changes on the backend:

- "Trait name" is changed to "Ontology term name";
- "Trait level" is changed to "Trait entitiy";
- "Trait attribute" is added.

### DIFF
--- a/src/views/trait/TraitsImport.vue
+++ b/src/views/trait/TraitsImport.vue
@@ -48,7 +48,7 @@
     <template v-if="state === ImportState.CHOOSE_FILE || state === ImportState.FILE_CHOSEN">
       <h1 class="title" v-if="showTitle">Import Traits</h1>
       <ImportInfoTemplateMessageBox v-bind:import-type-name="'Trait'"
-                                    v-bind:template-url="'https://cornell.box.com/shared/static/8sp0qvccpjotosiv8576tczeg09nnvao.xls'"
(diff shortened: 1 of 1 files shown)


timparsons:
Should columns such as 'Trait abbreviations' and 'Trait synonyms' be renamed to remove the 'Trait' from the column name?  Maybe this is better to be in a different card

src/components/trait/TraitImportTemplateMessageBox.vue
```diff
@@ -32,7 +32,7 @@
             <div class="level-item">
               <div class="has-text-dark has-text-centered is-size-7">
                 <!-- temporary link until the backend card is done -->
-                <a href="https://cornell.box.com/shared/static/8sp0qvccpjotosiv8576tczeg09nnvao.xls"
+                <a href="https://cornell.box.com/shared/static/u9vuob3vvu2cgwyoe26b2722u41b8f6o.xls"
```
nickpalladino, line 35: Had a question for @shawnyarnes on the api PR about Ontology term name so dependent on that

nickpalladino: Talked with @shawnyarnes and this will be addressed in BI-1059

timparsons (approved):

nickpalladino (approved):

PR #107, opened 2021-08-16
[BI-1079] Parse 400 errors for trait import errors

ctucker3:
Parses 400 errors on the front end to display a cleaner error message. Only 400 errors are parsed. If the error is not a 400 error, and not a `ValidationError`, a generic "An unknown error has occurred" message is shown. 

duplicate_columns
missing_columns
screencapture-192-168-1-113-8080-programs-9fbf27ad-b60d-4c13-8d2e-39e55784007d-traits-import-2021-08-16-16_19_50

### DIFF
--- a/src/components/file-import/FileSelectMessageBox.vue
+++ b/src/components/file-import/FileSelectMessageBox.vue
@@ -94,6 +94,7 @@
   import FileSelector from "@/components/file-import/FileSelector.vue";
   import {ValidationError} from "@/breeding-insight/model/errors/ValidationError";
   import { AlertTriangleIcon } from 'vue-feather-icons';
+  import {AxiosResponse} from "axios";
 
   @Component({
     components: {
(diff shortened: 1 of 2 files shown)


HMS17 (commented):
Should the bug tags on TAF scenarios BI-927 and BI-928 be removed as part of this JIRA since this change should fix them?

ctucker3:
> Should the bug tags on TAF scenarios BI-927 and BI-928 be removed as part of this JIRA since this change should fix them?

I imagine so. Would that be something Deb would do, or I would do? I'll also have to do it once this is merged, right?

nickpalladino (approved):

HMS17:
> > Should the bug tags on TAF scenarios BI-927 and BI-928 be removed as part of this JIRA since this change should fix them?
> 
> I imagine so. Would that be something Deb would do, or I would do? I'll also have to do it once this is merged, right?

I'm not sure we've determined whose particular responsibility it is so much as it should be updated, and yes, probably after merge but before the JIRA is marked as done? Going to approve this; just wanted to make sure we at least discussed keeping up the documentation.

HMS17 (approved):

PR #108, opened 2021-08-17
[BI-640] - Disable modification of user role for self

HMS17:
The current messages when a sys admin modifies their own role and no other fields are confusing. To simplify this, the field for modifying a user's role will be disabled when the user is the same as the sys admin.

**Changes:**
When user matches active user:
- In edit for System Users table, disabled Role field
- Only submitted change via promises for user info and not for user roles
- Use more specific user info success message 'User info (name/email/program) successfully updated'

Modified banner message logic for one promise succeeding and not the other to only fire when there are two promises

Linked to PR [taf/BI-640](https://github.com/Breeding-Insight/taf/pull/15)

### DIFF
--- a/src/components/admin/AdminUsersTable.vue
+++ b/src/components/admin/AdminUsersTable.vue
@@ -224,7 +224,7 @@
               v-bind:selected-id="editData.roleId"
               v-bind:field-name="'Role'"
               v-bind:empty-value-name="'No Role'"
-              :isDisabled="isCurrentUser(editData.id)"
+              v-bind:is-disabled="isCurrentUser(editData.id)"
             />
           </div>
         </div>
@@ -462,7 +462,7 @@ export default class AdminUsersTable extends Vue {
   }
 
   isCurrentUser(userId: string) {
-    if (this.activeUser!.id == userId){
+    if (this.activeUser!.id === userId){
       return true;
     } else {
       return false;
--- a/src/components/admin/AdminUsersTable.vue
+++ b/src/components/admin/AdminUsersTable.vue
@@ -224,6 +224,7 @@
               v-bind:selected-id="editData.roleId"
               v-bind:field-name="'Role'"
               v-bind:empty-value-name="'No Role'"
+              :isDisabled="isCurrentUser(editData.id)"
             />
           </div>
         </div>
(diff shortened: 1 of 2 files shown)


src/components/admin/AdminUsersTable.vue
```diff
@@ -442,6 +461,14 @@ export default class AdminUsersTable extends Vue {
     }
   }
 
+  isCurrentUser(userId: string) {
+    if (this.activeUser!.id == userId){
```
nickpalladino, line 465: Although not as much an issue in Typescript I think we've been sticking with === for equality checks

src/components/admin/AdminUsersTable.vue
```diff
@@ -224,6 +224,7 @@
               v-bind:selected-id="editData.roleId"
               v-bind:field-name="'Role'"
               v-bind:empty-value-name="'No Role'"
+              :isDisabled="isCurrentUser(editData.id)"
```
nickpalladino, line 227: For consistency I think we've been sticking with the v-bind syntax and kebab case prop name

nickpalladino (commented):
Do we need a card for updating TAF test cases?

src/components/admin/AdminUsersTable.vue
timparsons: +1

timparsons (approved):
Approving assuming suggested changes from Nick are addressed.

src/components/admin/AdminUsersTable.vue
HMS17: Resolved

src/components/admin/AdminUsersTable.vue
HMS17: Resolved. Interesting that with vue the @Prop listed can be camel case but the v-bind can have kebab case...

HMS17:
> Do we need a card for updating TAF test cases?

We would, yes. In UserManagementSysAd.feature we'd need possibly one extra scenario to test for for conditional disabling of the Roles field and then modifying BI-836 to no longer change role and not look for the role error banner. Should I make/pick up the card?

nickpalladino:
> > Do we need a card for updating TAF test cases?
> 
> We would, yes. In UserManagementSysAd.feature we'd need possibly one extra scenario to test for for conditional disabling of the Roles field and then modifying BI-836 to no longer change role and not look for the role error banner. Should I make/pick up the card?

Ya, I would say make the card. As far as working on it, not sure what the priorities are, could check with @timparsons

nickpalladino (approved):

HMS17:
> > > Do we need a card for updating TAF test cases?
> > 
> > 
> > We would, yes. In UserManagementSysAd.feature we'd need possibly one extra scenario to test for for conditional disabling of the Roles field and then modifying BI-836 to no longer change role and not look for the role error banner. Should I make/pick up the card?
> 
> Ya, I would say make the card. As far as working on it, not sure what the priorities are, could check with @timparsons

Added BI-1090. Not sure how high priority it is, but this PR will break scenario BI-836 so it's higher priority than a card that would just consist of adding a new scenario.

PR #109, opened 2021-08-17
[BI-975] - Prevent creating programs with same name

HMS17:
Added check and error message to prevent creating programs with the same name.

This error message takes priority over the "BrAPI URL specified is not supported" error.

Connected to PR [bi-api/BI-975](https://github.com/Breeding-Insight/bi-api/pull/101)


nickpalladino (approved):
Approved assuming any TAF cards needed exist

ctucker3 (approved):

PR #110, opened 2021-08-18
[BI-995] Ontology UI - Add New Term

dmeidlin:
- Additional required fields of Description, Entity, and Attribute were added to the new Ontology Term form.
- Client-side validations added for 12-char max limit on Name and 30-char max limit on Entity, Attribute, and Method Description.
- Server-side validations in place from BI-982 for all required fields: Name, Description, Entity, Attribute, Method Description, Method Class, and Scale.
- Archive status switch added to form.
- Form restyled to join inputs into logical groups: Ontology Term, Trait, Method, Scale, and Tags.
- Autocomplete inputs used for Entity, Attribute, and Method Description.
- Synonyms are automatically set and displayed to be Name and Full Name.
- Trait name is automatically set and displayed as Entity + Attribute.
- Method name is automatically set and displayed as Description + Class.


src/breeding-insight/model/ObservationVariable.ts
```diff
@@ -0,0 +1,71 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import {Trait} from "@/breeding-insight/model/Trait";
+import {Method} from "@/breeding-insight/model/Method";
+import {Scale} from "@/breeding-insight/model/Scale";
+
+export class ObservationVariable {
```
nickpalladino, line 22: Is this just for the future to change from Trait to ObservationVariable? Don't see this class being used anywhere. Is there a card to do that?

src/components/trait/forms/BaseTraitForm.vue
```diff
@@ -1,58 +1,137 @@
...
+  <div class="columns is-multiline is-gapless">
     <div class="column" v-bind:class="{'is-full': editFormat}">
+
+      <!--Ontology Term-->
+      <div class="columns">
+        <span class="column form-heading-toggle is-half">Ontology Term</span>
+        <div class="column is-half">
+          <input id="newTraitActiveToggle"
+                 type="checkbox"
+                 name="switchRtlExample"
```
nickpalladino, line 11: Could the switch be the rounded style to more closely match the mockup? Also maybe the name should be changed to something more appropriate.

src/components/trait/forms/BaseTraitForm.vue
```diff
@@ -1,58 +1,137 @@
...
-        />
-      </div>
-      <div class="sentence-input">
-        <p class="is-input-prepend mt-3">
-          using
-        </p>
-        <BasicSelectField
+      />
+      <BaseFieldWrapper class="new-form-content" fieldName="Method" show-label="true">
+        {{ methodName }}
```
nickpalladino, line 114: This is showing up as `undefined undefined` when the form is empty. Don't think it should show anything in that case.

src/components/trait/forms/BaseTraitForm.vue
```diff
@@ -1,58 +1,137 @@
...
+          v-bind:field-name="'Description'"
+          v-bind:field-help="'All unicode characters are accepted.'"
+          v-bind:placeholder="'Ontology Term Description'"
+          v-bind:show-label="true"
+          v-bind:server-validations="validationHandler.getValidation(0, TraitError.TraitDescription)"
+          v-on:input="trait.traitDescription = $event"
+      />
+
+      <BaseFieldWrapper class="new-form-content" fieldName="Synonyms" show-label="true">
+        {{ trait.synonyms ? trait.synonyms.join(' ') : '' }}
```
nickpalladino, line 52: could these be comma separated or shown like tags maybe so it's easy to distinguish between each synonym?

src/components/trait/forms/BaseTraitForm.vue
```diff
@@ -1,58 +1,137 @@
...
       <BasicInputField
-        v-bind:value="trait.traitName"
-        v-bind:field-name="'Trait name'"
-        v-bind:field-help="'All unicode characters are accepted.'"
-        v-bind:placeholder="'Trait Name'"
-        v-bind:server-validations="validationHandler.getValidation(0, TraitError.TraitName)"
-        v-on:input="trait.traitName = $event"
+          class="new-form-content"
+          v-bind:value="trait.observationVariableName"
+          v-bind:field-name="'Name'"
```
nickpalladino, line 22: After saving a trait, the name is empty in the trait list table? Is this being handled in Heather's card?

src/components/trait/forms/BaseTraitForm.vue
```diff
@@ -1,58 +1,137 @@
 <template>
-  <div class="columns is-multiline">
+  <div class="columns is-multiline is-gapless">
     <div class="column" v-bind:class="{'is-full': editFormat}">
+
+      <!--Ontology Term-->
+      <div class="columns">
```
nickpalladino, line 6: How closely do we want the new trait form to match the mockup @shawnyarnes? Looks quite a bit different at this point.
Screenshot from 2021-09-13 11-38-56

src/components/trait/TraitListsTable.vue
```diff
@@ -439,6 +463,20 @@ export default class TraitTable extends Vue {
     this.newTraitActive = false;
   }
 
+  async getAttributesEntitiesDescriptions() {
```
nickpalladino, line 466: I don't seem to be getting previously saved values showing up in the autocomplete box.

src/breeding-insight/service/TraitService.ts
```diff
@@ -143,4 +143,24 @@ export class TraitService {
...
+
+  static async getAttributesEntitiesDescriptions(programId: string): Promise<[string[], string[], string[]]> {
+    if (programId) {
+      try {
+        const [ traits, metadata ] = await TraitService.getAll(programId, new PaginationQuery(1,50,true), true);
+        let attributes: Set<string> = new Set();
+        let entities: Set<string> = new Set();
+        let descriptions: Set<string> = new Set();
+        for (const trait of traits) {
+          if(trait.attribute) attributes.add(trait.attribute);
```
nickpalladino, line 155: Are these all case standardized on the backend so there won't be duplicates with different cases?

src/components/trait/forms/BaseTraitForm.vue
dmeidlin: Yeah, that happened after after I changed the code to fix a TS error and I didn't catch it before issuing the PR review, but I fixed it in the last commit that added an asterisk to required fields.

src/components/trait/forms/BaseTraitForm.vue
dmeidlin: Yes, that's handled in BI-994.

src/components/trait/forms/BaseTraitForm.vue
dmeidlin: I just showed it to Shawn this morning and she seemed pleased with it. The only difference that we talked about was adding an asterisk after required field labels.  It's a bit of a challenge matching the wireframe exactly and balancing the scope of the card. For example, it's hard to take up the space between the inputs because of the helper text and additional padding of the div wrappers for optional field-error messages.  There could certainly be improvements to the styling but I thought that would be something to focus on after MVP, anyway.

nickpalladino: sounds good

src/components/trait/TraitListsTable.vue
dmeidlin: If you don't have the changes to bi-api from BI-982, then you won't see those values. The frontend is calling the traits endpoint, and those fields were added to the model in that card.

nickpalladino: I have the api 995 branch so should have all that

src/breeding-insight/service/TraitService.ts
dmeidlin: I just tested all autocompletes(entity, attribute, method description) and only lower case text is returned in the dropdowns:no duplicates.

src/breeding-insight/model/ObservationVariable.ts
dmeidlin: Yeah, I thought in the beginning I would need it, but it wasn't necessary.  It will be necessary in the future, though, so it made sense to me to leave it in, but I can see the case for leaving out extraneous code, too. So, I'm open to whatever you want to do, either way.

nickpalladino: maybe just put a comment on it with a TODO to refactor Trait into that and if we have a card to do that in created.

src/components/trait/forms/BaseTraitForm.vue
dmeidlin: Rounding it is no problem, and, yes, I'll change the name to something appropriate: good catch!

Done.

src/components/trait/forms/BaseTraitForm.vue
dmeidlin: Yep. I'll add the comma.

Done.

src/components/trait/TraitListsTable.vue
dmeidlin: Ok, thanks. The issue was that the method that fetches the autocomplete options was called whenever the `TraitsListTable.vue` component was mounted, but not after the `save` button was pressed. I ficed it and pushed it up, so iy should work for you, now. Also, because the autocomplete options weren't updating I tricked myself into thinking that only lowercase options were displayed. There are duplicates in the options and the backend needs to check for the case.

nickpalladino: Is the checking something the backend already does or needs to be a new card?

nickpalladino (approved):

src/breeding-insight/model/ObservationVariable.ts
dmeidlin: Done.

src/breeding-insight/service/TraitService.ts
dmeidlin: @shawnyarnes thought of an edge case in which a common attribute such as pH would be used. If all attributes were stored as lower case than the client would need to format edge cases like this. Also, she would like to have a prompt appear whenever the user tries to input a custom string instead of an autocomplete term which would ask the user to confirm that they want to proceed, possibly showing similar existing terms already stored.  That could address the issue of a user inadvertently adding duplicate terms that only differ by case.

I propose creating a new card for the prompt and leaving the autocomplete as is for now. Thoughts?

timparsons: Seems reasonable to me, but agreed that it should be in a different card

timparsons (approved):

PR #111, opened 2021-08-23
[BI-380] Update sidebar layout to stay in place and be independently scrollable from page scroll

nickpalladino:
Make sure you run `npm install` because vue-feather-icons was updated to get the logout icon.

- Made header fixed to top and made smaller
- Moved user info and logout button to user menu in header from main content area so it's always visible
- Made copyright date dynamic based on local time
- Made sidebar menu independently scrollable from main content
- Keep version info at bottom of sidebar menu
  - Ran out of time so it's at the bottom of content but not fixed, could be better

Cards created:
- [BI-1099 - Mobile Friendly Improvements](https://breedinginsight.atlassian.net/browse/BI-1099?atlOrigin=eyJpIjoiMWE4ZjFmOGRiNWI4NDFjZTg0ZmJkOWMxMmJjMjQyOGUiLCJwIjoiaiJ9)
- [BI-1100 - TAF Updates for Sidebar Layout Changes](https://breedinginsight.atlassian.net/browse/BI-1100?atlOrigin=eyJpIjoiM2I2YTA1YWE0NzZkNDlkMDhjZGI1YmM2MmEzMmNmYzgiLCJwIjoiaiJ9)


HMS17 (approved):

src/components/layouts/BaseSideBarLayout.vue
```diff
@@ -109,23 +93,34 @@ import { Component, Prop, Vue, Watch } from 'vue-property-decorator';
 import { MenuIcon } from 'vue-feather-icons';
 import { SandboxMode } from '@/util/config';
 import VersionInfo from '@/components/layouts/VersionInfo.vue';
-
+import Footer from "@/components/layouts/Footer.vue";
```
timparsons, line 96: I don't think this is used in the component

src/assets/scss/main.scss
```diff
@@ -177,6 +177,10 @@ footer {
   flex-shrink:0;
 }
 
+.main-footer {
```
timparsons, line 180: I'm wondering if it'd be better to modify the structure of the HTML to have the main footer be a part of the main body, rather than modifying the footer to look like it's a part of the main body via CSS.

src/assets/scss/main.scss
```diff
@@ -466,7 +493,6 @@ a.is-underlined {
 }
 
 #versionInfo {
```
timparsons, line 495: Could some margin padding be added to the top of the version info in the menu footer?  It feels a little cramped with the menu items

timparsons (changes requested):
A couple of small things.  Everything else looks great!

src/components/layouts/BaseSideBarLayout.vue
nickpalladino: ah ya thanks that was from some early messing around that was left over. Removed and pushed changes.

src/assets/scss/main.scss
nickpalladino: ya that was the first approach I tried but had some display issues and got tired of messing with it so switched to keeping the footer and modifying it based on the layout. Germinate does a similar thing as well, has a footer and uses margin to move it around. I guess I'd rather keep it then spend time messing around trying to get it to work the other way. I'm not opposed to reconsidering in the future though.

src/assets/scss/main.scss
nickpalladino: Added some padding and pushed changes. I'll create another card as well to make the version fixed to the bottom rather than scroll.

timparsons (approved):

PR #112, opened 2021-08-25
[BI-1047] - Convert expandable datatables to Buefy

HMS17:
Following [BI-895](https://github.com/Breeding-Insight/taf/pull/9/), converting other expandable datatables to use Buefy ExpandableTable component.

Converted following tables to use ExpandableTable
- Admin Users Table
- Admin Programs Table
- Program Locations Table
- Trials Table
- Observations List Table
 
Removed ExpandableRowTable.vue and associate unit tests
Modified Program Users Table to use the heading "Role" instead of "Roles"

Linked to PR [taf/BI-1047](https://github.com/Breeding-Insight/taf/pull/17)


src/components/admin/AdminProgramsTable.vue
```diff
@@ -213,9 +209,10 @@
 @Component({
   mixins: [validationMixin],
   components: {
-    EmtpyTableMessage,
+    ExpandableTable, EmtpyTableMessage,
     NewDataForm, WarningModal, PlusCircleIcon,
-    ExpandableRowTable, TableColumn, BasicInputField, BasicSelectField
+    TableColumn, BasicInputField, BasicSelectField,
+    ChevronRightIcon, ChevronDownIcon
```
nickpalladino, line 215: Think some of these icon imports could be cleaned up, don't seem to be needed here.

src/components/admin/AdminProgramsTable.vue
```diff
@@ -61,6 +61,8 @@
       </span>
     </button>
 
+    <div class="is-clearfix"></div>
```
nickpalladino, line 64: I noticed this being used in a bunch of the table pages but not for trials or observations. Not sure what it's used for?

HMS17: I'm not quite sure. I'm aware clearfix can be used to help fix broken layouts and I was patterning my changes off the PR for converting the Program Users table, which also included adding this.

src/components/admin/AdminProgramsTable.vue
HMS17: Which icons are you speaking of? Since I do think the Chevron icons are used (when sorting columns).

nickpalladino: ya chevron, I think those should be in the buefy icon pack so doesn't seem like they'd be needed in this component

src/components/admin/AdminProgramsTable.vue
nickpalladino: ok, I guess if it's serving some purpose it should probably be in all of the tables.

src/components/admin/AdminProgramsTable.vue
HMS17: That was another change where I was mirroring the Program Users table conversion.

nickpalladino: ya, just wondering if that slipped through when that first one was reviewed. It's not a big deal, I'll approve either way.

src/components/admin/AdminProgramsTable.vue
nickpalladino: @timparsons, did you add this in there?

timparsons: Yep, I had to add that because the 'New ' button above the table has the `is-pulled-right` class, and this was causing an overlay to cover up the action buttons in the table. Doing the clearfix between the button and the table resolved that issue.

dmeidlin (approved):

src/components/trials/TrialsTable.vue
```diff
@@ -38,7 +38,7 @@
     <button
       v-if="$ability.can('create', 'Trial')"
       data-testid="newDataForm"
-      v-show="!newTrialActive & trials.length > 0"
+      v-show="!newTrialActive"
       class="button is-primary has-text-weight-bold is-pulled-right"
       v-on:click="newTrialActive = true"
     >
```
nickpalladino, line 44: think this should have the clearfix between the button and table right?

nickpalladino (changes requested):

src/components/trials/TrialsTable.vue
HMS17: Good catch; fixed

nickpalladino (approved):

PR #116, opened 2021-09-03
[BI-997 BI-1061] Update Trait Import

HMS17:
Ontology Changes 
- Replace "Trait" with "Ontology" in Trait Import
- Alter table columns in Trait Import according to specifications

Linked to PR [taf/BI-997](https://github.com/Breeding-Insight/taf/pull/21)


src/views/trait/TraitsImport.vue
```diff
@@ -46,8 +46,8 @@
     </WarningModal>
```
nickpalladino, line 46: Looks like the WarningModal still has trait terminology in the text that should be updated to ontology

HMS17: I don't believe the warning modal was part of either JIRA? Admittedly it is something that should be changed.

nickpalladino: Not sure if it was specifically called out but I think it meets the intent of the card and it's a quick change so might as well do it while you're in there.

HMS17: Noted. I will have to potentially change TAF as well if there are tests. Also heads up that I put this in a new PR (https://github.com/Breeding-Insight/bi-web/pull/125) because I thought the build failures are likely a consequence of all the messy merges. Apparently that is not the case but at least I ruled it out and have a cleaner PR.

nickpalladino: ok I'll look at that one

HMS17:
Moved changes to cleaner PR https://github.com/Breeding-Insight/bi-web/pull/125

PR #119, opened 2021-09-09
[BI-994] Ontology UI - Review Existing Terms

HMS17:
Modified sidebar menu to use Ontology instead of Traits
Changed existing traits list to an ontology list and changed table and associated buttons to match desired UI


HMS17:
Closing and reopening a cleaner PR

PR #120, opened 2021-09-14
BI-760 (Notification handling improvements)

davedrp:
using the vuex store to hold the state of the notify-components


HMS17 (approved):

src/components/notifications/ErrorNotification.vue
```diff
@@ -30,17 +30,33 @@
...
-  public msg : string = '';
+  // 'active' and 'msg' should stay private.
+  // (use the getters and mutators in the vuex store
+  //  to modify these values)
+  private active: boolean = false;
+  private msg: string = '';
 
   private bannerTextClass: string = "banner-text";
+
+  @Watch('$store.state.errorNotificationActive', {immediate: true})
```
nickpalladino, line 49: I think you could use the vuex getters here instead of watch if you wanted to have a bit less code because I see you already defined them in `src/store/index.ts`

src/components/notifications/ErrorNotification.vue
```diff
@@ -16,8 +16,8 @@
   -->
 
 <template>
-  <b-notification type="is-danger" v-bind:active.sync="active" aria-close-label="Close Notification" 
-    role="alert">
+  <b-notification type="is-danger" v-bind:active.sync="active" aria-close-label="Close Notification"
+    role="alert" v-on:close="onClose">
     <div class="columns has-vertical-align-middle">
```
nickpalladino, line 21: Not related to any of your changes but I noticed the error notification takes up more vertical space than the success and info message banners so the sizes aren't the same. Would you be willing to update this template layout to match the others so the sizes are consistent?

src/components/forms/DataForm.vue
```diff
@@ -110,9 +116,17 @@ export default class DataForm extends Vue {
     if (this.$v.record) {
       this.$v.record.$reset();
     }
+    this.closeNotifications();
     this.$emit('cancel');
   }
 
+  closeNotifications() {
+    this.$store.commit(DEACTIVATE_ERROR_NOTIFICATION);
```
nickpalladino, line 124: Maybe this can be moved to a vuex action then it's only in one spot and can be reused.

package.json
```diff
@@ -22,6 +22,7 @@
     "@types/promise.allsettled": "^1.0.3",
     "@types/valid-url": "^1.0.3",
     "@types/vuelidate": "^0.7.10",
+    "@vue/cli": "^4.5.13",
```
nickpalladino, line 25: What was this needed for?

nickpalladino (commented):
It's a little unclear to me from the jira card the exact behavior of notifications closing related to form updates. One example is I created a program and got a success notification. I then clicked edit on one of the table rows and then clicked cancel and the notification closed. technically this was a different form than the one that produced the notification. Not sure if that matters? @DebWeigand @timparsons

timparsons:
> It's a little unclear to me from the jira card the exact behavior of notifications closing related to form updates. One example is I created a program and got a success notification. I then clicked edit on one of the table rows and then clicked cancel and the notification closed. technically this was a different form than the one that produced the notification. Not sure if that matters? @DebWeigand @timparsons

@nickpalladino I think if anything, any current notifications should be cleared when a new form is opened.  @shawnyarnes thoughts?

package.json
davedrp: This is not needed and has been removed

src/components/notifications/ErrorNotification.vue
davedrp: I'm leery of making the changes at this time.  The error notification banner is structured a little differently than the other notifications and has different properties (for example, a long message will wrap in the error notification, but not in the others).  I am concerned that a change could have unintended consequences.

src/components/forms/DataForm.vue
davedrp: For now, I'm going to leave this as is.

PR #124, opened 2021-09-16
[BI-994] Ontology UI - Review Existing Terms

HMS17:
Modified sidebar menu to use Ontology instead of Traits
Changed existing traits list to an ontology list and changed table and associated buttons to match desired UI

Linked to PR [taf/BI-994](https://github.com/Breeding-Insight/taf/pull/26)

### DIFF
--- a/src/components/trait/TraitListsTable.vue
+++ b/src/components/trait/TraitListsTable.vue
@@ -45,7 +45,7 @@
       </div>              
     </WarningModal>
 
-    <div class="columns has-text-right mb-0">
+    <div class="columns has-text-right mb-0 buttons">
       <div class="column">
         <button
             v-if="$ability.can('create', 'Trait')"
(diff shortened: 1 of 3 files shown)


timparsons (approved):

src/components/trait/TraitListsTable.vue
```diff
@@ -158,15 +179,15 @@
       <template v-slot:emptyMessage>
         <EmptyTableMessage
             v-bind:button-view-toggle="!newTraitActive"
-            v-bind:button-text="'New Trait'"
+            v-bind:button-text="'New Term'"
```
nickpalladino, line 182: I'm noticing some difference in terminology between using ontology and ontology terms that seems like it could be more consistent. I'm ok with it going through for now as-is.

nickpalladino (approved):

src/components/trait/TraitListsTable.vue
HMS17: I talked to @shawnyarnes about phrasing and when to use "Ontology" vs "Ontology Term"

nickpalladino: ya some specific spots that I'm thinking are like `Archived Ontology` seems like it should be `Archived Ontology Terms` because we aren't really archiving the entire ontology (set of terms). Things like that.

HMS17: Makes sense. Might be a good conversation to have with Shawn then.

PR #129, opened 2021-09-30
[BI-1132] Keep the Left Nav Menu from overlapping with the header while in public sanbox mode

dmeidlin:
The overflow of the left navbar into the header resulted from the header height changing depending on the presence/absence of the sandbox mode notification badge.

The header height is now fixed and sandbox badge resized to fit the fixed header height.


src/components/layouts/BaseSideBarLayout.vue
```diff
@@ -43,12 +43,10 @@
           <div v-if="sandboxConfig !== ''" class="level-item">
             <div v-bind:class="{'notification is-warning px-5 has-text-centered': sandboxConfig === SandboxMode.Public,
                                 'notification is-info px-5 has-text-centered': sandboxConfig === SandboxMode.Coordinator}">
-              <p class="title is-size-4">Sandbox</p>
-              <p>
-                <a href="#" v-on:click="$showCollectorDialog()" v-bind:class="{'has-text-link': sandboxConfig === SandboxMode.Public,
+              <span class="title is-size-4">Sandbox </span>
+              <a href="#" v-on:click="$showCollectorDialog()" v-bind:class="{'has-text-link': sandboxConfig === SandboxMode.Public,
```
nickpalladino, line 47: Would it be possible to keep the previous vertical layout with feedback below the sandbox text and taking up the whole header height but not pushing the header down?

dmeidlin: I changed the button text to be vertical and centered. Have a look and see if you like it. The header padding on top and bottom has to be smaller than the other header elements, so that visually isn't consistent but can't be avoided.  The previous approach with horizontal text had consistent header padding top and bottom; that's the only significant difference I see.

nickpalladino (approved):

src/components/layouts/BaseSideBarLayout.vue
davedrp: look good

PR #144, opened 2021-11-09
[BI-503] Add a loading indicator for Confirm Traits

HMS17:
- For both Germplasm and Trait Import, on clicking the "Confirm" button, the button will be disabled and a loading indicator will appear.
- (Changes were applied to Germplasm Import because both Traits and Germplasm Import use the  ConfirmImportMessageBox.vue component with the aforementioned "Confirm" button)

### DIFF
--- a/src/views/import/ImportTemplate.vue
+++ b/src/views/import/ImportTemplate.vue
@@ -356,7 +356,6 @@ export default class ImportTemplate extends ProgramsBase {
   }
 
   async confirm() {
-    this.confirmImportState.bus.$emit(DataFormEventBusHandler.SAVE_STARTED_EVENT);
     const name = this.activeProgram && this.activeProgram.name ? this.activeProgram.name : 'the program';
     try {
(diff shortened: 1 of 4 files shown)


src/views/import/ImportTemplate.vue
```diff
@@ -350,7 +356,7 @@ export default class ImportTemplate extends ProgramsBase {
   }
 
   async confirm() {
-
+    this.confirmImportState.bus.$emit(DataFormEventBusHandler.SAVE_STARTED_EVENT);
```
timparsons, line 359: I would recommend putting this emit in the ConfirmImportMessageBox.vue `confirm` method

timparsons (changes requested):

src/views/import/ImportTemplate.vue
HMS17: Changed!

timparsons (approved):

dmeidlin (approved):

PR #146, opened 2021-11-11
[BI-623] Update UI OntologyTables implementation to handle sort events

dmeidlin:
The service and dao for Traits was changed to pass the query params used for sorting.
Vuex was used to store sort control state for each table. This was necessary to ensure that the display for a table is preserved when a user selects the other tab in the interface or when the user navigates to another list table.

The approach used here will be replicated for tables in the remaining sub tasks of [BI-618](https://breedinginsight.atlassian.net/jira/software/c/projects/BI/boards/1?selectedIssue=BI-618).

Backend sorting was implemented for the following tables:
admin users,
programs,
program users,
locations,
active ontology,
archived ontology,
trait import preview.

The ontology and trait import preview table sorting was styled to match that used in the other Buefy tables.

The bugfix for [BI-1115](https://breedinginsight.atlassian.net/browse/BI-1115) was also merged in, so backend pagination is being used for the program users table.

### DIFF
--- a/src/breeding-insight/model/Sort.ts
+++ b/src/breeding-insight/model/Sort.ts
@@ -51,11 +51,11 @@ export enum OntologySortField {
 
 export class OntologySort {
   field: OntologySortField;
-  flag: boolean;
+  order: SortOrder;
 
-  constructor(field: OntologySortField, flag: boolean) {
+  constructor(field: OntologySortField, order: SortOrder) {
     this.field = field;
-    this.flag = flag;
+    this.order = order;
   }
 }
 
--- a/src/breeding-insight/model/Sort.ts
+++ b/src/breeding-insight/model/Sort.ts
@@ -1,0 +1,28 @@
+/*
+ * See the NOTICE file distributed with this work for additional information
+ * regarding copyright ownership.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+export enum SortOrder {
+  Ascending = 'ASC',
+  Descending = 'DESC'
+}
+
+export enum TraitSortField {
+  Name = 'name',
+  MethodDescription = 'methodDescription',
+  ScaleClass = 'scaleClass',
+  ScaleName = 'scaleName'
+}
--- a/src/breeding-insight/model/Sort.ts
+++ b/src/breeding-insight/model/Sort.ts
@@ -26,3 +26,79 @@ export enum TraitSortField {
   ScaleClass = 'scaleClass',
   ScaleName = 'scaleName'
 }
+
+// ontology
+export enum OntologySortField {
+  Name = 'name',
+  MethodDescription = 'methodDescription',
+  ScaleClass = 'scaleClass',
+  ScaleName = 'scaleName'
+}
+
+export class OntologySort {
+  field: OntologySortField;
+  order: SortOrder;
+
+  constructor(field: OntologySortField, order: SortOrder) {
+    this.field = field;
+    this.order = order;
+  }
+}
+
+// location
+export enum LocationSortField {
+  Name = 'name',
+  Abbreviation = 'abbreviation',
+  Slope = 'slope'
+}
+
+export class LocationSort {
+  field: LocationSortField;
+  order: SortOrder;
+
+  constructor(field: LocationSortField, order: SortOrder) {
+    this.field = field;
+    this.order = order;
+  }
+}
+
+// user
+export enum UserSortField {
+  Name = 'name',
+  Email = 'email',
+  Roles = 'roles',
+  Active = 'active'
+}
+
+export class UserSort {
+  field: UserSortField;
+  order: SortOrder;
+
+  constructor(field: UserSortField, order: SortOrder) {
+    this.field = field;
+    this.order = order;
+  }
+}
+
+// program
+export enum ProgramSortField {
+  Name = 'name',
+  Abbreviation = 'abbreviation',
+  Objective = 'objective',
+  DocumentationUrl = 'documentationUrl',
+  Active = 'active',
+  BrapiUrl = 'brapiUrl',
+  NumUsers = 'numUsers',
+  SpeciesId = 'speciesId',
+  SpeciesName = 'speciesName'
+}
+
+export class ProgramSort {
+  field: ProgramSortField;
+  order: SortOrder;
+
+  constructor(field: ProgramSortField, order: SortOrder) {
+    this.field = field;
+    this.order = order;
+  }
+}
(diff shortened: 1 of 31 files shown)


src/breeding-insight/model/Sort.ts
```diff
@@ -0,0 +1,28 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+export enum SortOrder {
+  Ascending = 'ASC',
+  Descending = 'DESC'
+}
+
+export enum TraitSortField {
```
timparsons, line 23: I assume the intent of this file is to add more enums for sorting fields of different types?

src/breeding-insight/service/TraitService.ts
```diff
@@ -120,25 +120,26 @@ export class TraitService {
...
-  static getFilteredTraits(programId: string, paginationQuery?: PaginationQuery, full?: boolean, filters?: TraitFilter[]): Promise<[Trait[], Metadata]> {
+  static getFilteredTraits(programId: string, paginationQuery?: PaginationQuery, full?: boolean, filters?: TraitFilter[], sortField?: TraitSortField, sortOrder?: SortOrder): Promise<[Trait[], Metadata]> {
     return new Promise<[Trait[], Metadata]>(((resolve, reject) => {
 
       if (paginationQuery === undefined) {
         paginationQuery = new PaginationQuery(0, 0, true);
       }
 
+      if (sortField === undefined) sortField = TraitSortField.Name;
+      if (sortOrder === undefined) sortOrder = SortOrder.Ascending;
```
timparsons, line 131: If you wanted, you could set these default values in the parameters defined in the function signature.

src/store/sorting/getters.ts
```diff
@@ -0,0 +1,57 @@
...
+    },
+    activeOntMethodSortOrder(state: SortState): boolean {
+        return state.activeOntMethodSortOrder;
+    },
+    activeOntScaleClassSortOrder(state: SortState): boolean {
+        return state.activeOntScaleClassSortOrder;
+    },
+    activeOntUnitSortOrder(state: SortState): boolean {
+        return state.activeOntUnitSortOrder;
+    },
```
timparsons, line 39: Why do the different columns have to be their own state for sort order?  Looks like there's one state for which field to sort on, so wondering if you could have one state for sort order.

timparsons (changes requested):

src/store/sorting/getters.ts
dmeidlin: It's set up that way as a convenience for the user. If there is a preference to sort in different directions on different columns, you can switch back and forth between columns with one click, but that wasn't really a prime design consideration.  It was just easy enough to make it with the most general controls for the user.

src/breeding-insight/model/Sort.ts
dmeidlin: Yes, that's right.

src/store/sorting/getters.ts
dmeidlin: There isn't a way to have independent sort order states for each field using the Buefy sorting api, so I will change the ontology table to use a single sort order state as well to be consistent with the Buefy tables.

ctucker3 (changes requested):
Code looks good. I have a few comments below on observations from running the code. I think only 1, 2 and 3 need your attention. 4 and 5 are probably more for @timparsons. 

1. Can we remove the underline on the arrow? See image below for traits list. Also, this icon is different from the other sorting icons, is this intentional?
underlined_arrow

2. For the traits table, would it be possible to add an underline on the column like buefy does when you hover over it so the user knows they can interact with that column? 

3. For BI-618 sorting should be added to the trait confirmation page, but I don't see any sorting ability on that one. 

no_sorting_trait_confirm

4. Looks like we need to fix the backend sorting. t11 should come before t10 here. (I'm not suggesting to do that in this card). 

out_of_order_sorting

5. I don't think this is a bug from your card, but I'll mark the bug here, just in case, since I see vuex changes in your code. When I create a new program, I get a success message. When I close that success message, it closes, but I get an error in the console:

```[Vue warn]: Computed property "isSuccessNotificationActive" was assigned to but it has no setter.```

File: src/breeding-insight/service/TraitService.ts [hunk 4]
```diff
@@ -120,25 +120,20 @@ export class TraitService {
...
-      }
-
       if (programId) {
-        TraitDAO.getFilteredTraits(programId, paginationQuery, full, filters).then((biResponse) => {
+        TraitDAO.getFilteredTraits(programId, paginationQuery, full, sort, filters).then((biResponse) => {
 
           let traits: Trait[] = [];
 
           if (biResponse.result.data) {
             //TODO: Remove when backend default sorting is implemented
```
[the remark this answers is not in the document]
dmeidlin: It looks like you ran under a prior commit; points 1 and 2 aren't an issue if you pull the latest commit 5f7ebfd. As for point 3, BI-618 calls for traits import preview to be sorted, which is what I changed and you can see it when navigating Ontology > Import Ontology, but the Ontology component is what's called when using Import File > Ontology, so you're right I should update that table as well. As for point 5, yeah I noticed the same while working on this and and agree that a bugfix card should be created.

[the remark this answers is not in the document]
dmeidlin: Actually, all of the ontology importing routes uses the same component, `TraitsImportTable.vue`, so there is nothing additional for me to do. Point 3 should also be fine once you pull the latest.

src/breeding-insight/model/Sort.ts
```diff
@@ -0,0 +1,118 @@
...
+export enum OntologySortField {
+  Name = 'name',
+  MethodDescription = 'methodDescription',
+  ScaleClass = 'scaleClass',
+  ScaleName = 'scaleName'
+}
+
+export class OntologySort {
+  field: OntologySortField;
+  flag: boolean;
```
timparsons, line 54: I saw that this boolean indicated if it was ASC or DESC order, could the name of this field be either `ascending` or `descending` (whatever `true` means) OR change this field to be of type `SortOrder`?

timparsons (changes requested):

timparsons:
@ctucker3 

> Looks like we need to fix the backend sorting. t11 should come before t10 here. (I'm not suggesting to do that in this card).

Yeah, a new card to handle this seems appropriate

src/breeding-insight/service/TraitService.ts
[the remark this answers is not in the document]
ctucker3: Oops, sorry about that! Just retested, 1 isn't an issue any more. 

2 still seems to be an issue, when I hover over a sortable column like `Method` that isn't currently being sorted, the pointer changes, but the header is not underlined like it usually is in the buefy table. The programs table displays the correct behavior. 

For 3, my initial comment isn't an issue anymore. However, on the trait import table, I can only switch to sort descending if I click directly on the icon, clicking on the header does not work. This seems to be the case for the Ontology list table as well. 

Also, for the TODO, could you check again? I still see it in the PR.

src/breeding-insight/model/Sort.ts
dmeidlin: Good idea, I changed the field to `order: SortOrder`.

timparsons (approved):

src/breeding-insight/service/TraitService.ts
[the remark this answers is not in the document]
dmeidlin: Thanks for pointing out the hover state; I never noticed that on Buefy tables before! The ontology tables behave the same as the Buefy tables, now.

ctucker3 (approved):

PR #147, opened 2021-11-11
[BI-1201] Add 'Program Key' field to Program Create

HMS17:
- Added field on admin program create for user to input a string for the program key
- This field is not present for updating programs.
- Added handling for checking if key is already in use
- Added front-end validations for key

Linked to PR [bi-api/BI-1201](https://github.com/Breeding-Insight/bi-api/pull/114)

### DIFF
--- a/src/breeding-insight/service/ProgramService.ts
+++ b/src/breeding-insight/service/ProgramService.ts
@@ -46,6 +46,7 @@ export class ProgramService {
           } else {
             error['errorMessage'] = this.errorCreatingProgram;
           }
+          //TODO handle duplicate program key
           reject(error);
         });
       }
@@ -105,7 +106,7 @@ export class ProgramService {
 
         // Parse our programs into the vue programs param
         programs = biResponse.result.data.map((program: any) => {
-          return new Program(program.id, program.name, program.species.id, program.numUsers, program.brapiUrl);
+          return new Program(program.id, program.name, program.species.id, program.numUsers, program.brapiUrl, program.key);
         });
 
         //TODO: Remove when backend pagination is implemented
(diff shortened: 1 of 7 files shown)


src/breeding-insight/service/ProgramService.ts
```diff
@@ -39,8 +40,10 @@ export class ProgramService {
           resolve(newProgram);
 
         }).catch((error) => {
-          if (error.response && error.response.status === 409) {
+          if (error.response && error.response.status === 409 && error.response.statusText === "Program name already in use") {
             error['errorMessage'] = this.duplicateProgramName;
+          } else if (error.response && error.response.status === 409 && error.response.statusText === "Program key already in use") {
+            error['errorMessage'] = this.duplicateProgramKey;
```
timparsons, line 46: This could be brittle if the text changes at all.  It may be worth looking into returning an enum for the status and evaluating that instead of a display string.

timparsons (changes requested):

src/breeding-insight/service/ProgramService.ts
HMS17: That makes sense. I was using the statusText as an imperfect solution since this change broke the assumption that there'd be only one error type for each status code. Are there examples of returning an enum elsewhere in the codebase?

timparsons: @HMS17 At the moment, no, there are no examples in the existing code for this.  Happy to talk through ways in which this could be done!

timparsons (approved):

src/breeding-insight/service/ProgramService.ts
HMS17: Implemented.

ctucker3 (approved):

PR #153, opened 2021-11-23
[BI-1115-2] Fix sortting and switch to backend paging

dmeidlin:
The bugfix for [BI-1115](https://github.com/Breeding-Insight/bi-web/pull/128) that enables backend paging with backend sorting was applied to the following tables:
programs,
admin users,
ontology active,
ontology archived,
ontology import preview,
program users, and
program locations.


HMS17 (approved):

ctucker3 (changes requested):
BI-1115 needs to be brought up to date with develop before I can test this. Develop branch for BI-API doesn't seem to work with this branch when creating a program. Looks like this branch is missing the program key, so that could be it?

ctucker3 (changes requested):
I noticed a bug with the programs table. Below are the steps to reproduce it:

1. Add a single program
2. Change the "Amount per page" dropdown to something other than its default 50
3. Create a new program
4. You'll notice the new program doesn't show up in the list. But if you refresh the page, it shows up.

ctucker3 (approved):
This bug is being worked on in another card:
- If I change the pagination, the sorting order change doesn't work anymore. The programs table doesn't seem to be changing the order of the rows when I switch for DESC to ASC. The proper request is send to the backend and the response is sorted properly, but the data in the table doesn't change.

I also noticed that the backend doesn't sort accurately for strings with numbers in them. If I have Test9 and Test11, when sorting descending order, Test9 will show up at the top of the lists and Test11 on the bottom with entries like Test1 and Test2. This might want to be captured in a card. 

Everything seems to work fine other than the bugs causes by the pagination. Approving since the bugs with this card are being done in other cards.

PR #154, opened 2021-12-03
[BI-1249] Tags - Saving after <Cancel>

davedrp:
[BI-1249](https://breedinginsight.atlassian.net/browse/BI-1249)

I noticed that Synonyms was having the same problem for the same reason, so I fixed that as well.

### DIFF
--- a/src/breeding-insight/model/Trait.ts
+++ b/src/breeding-insight/model/Trait.ts
@@ -87,8 +87,17 @@ export class Trait {
   }
 
   static assign(trait: Trait): Trait {
+    // if arrays exist in trait, then COPY them
+    let tags;
+    if (trait.tags){
+       tags = Array.from(trait.tags);
+    }
+    let synonyms;
+    if (trait.synonyms){
+      synonyms = Array.from(trait.synonyms);
+    }
     return new Trait(trait.id, trait.traitName, trait.observationVariableName, trait.programObservationLevel, trait.entity, trait.attribute,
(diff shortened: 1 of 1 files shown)


src/breeding-insight/model/Trait.ts
```diff
@@ -87,8 +87,17 @@ export class Trait {
...
   static assign(trait: Trait): Trait {
+    // if arrays exist in trait, then COPY them
+    let tags;
+    if (trait.tags){
+       tags = Array.from(trait.tags);
+    }
+    let synonyms;
+    if (trait.synonyms){
+      synonyms = Array.from(trait.synonyms);
+    }
```
ctucker3, line 98: I'm wondering if these array reassignments would be better done in the Trait constructor? With this method, I think we still run the danger of linking trait tags if we ever assign a new trait the values of another trait tags.

src/breeding-insight/model/Trait.ts
```diff
@@ -126,6 +136,8 @@ export class Trait {
         (this.method && this.method.equals(trait.method)) ||
         (!this.method && !trait.method)
       ) &&
+        (JSON.stringify(this.tags) === JSON.stringify(trait.tags)) &&
+        (JSON.stringify(this.synonyms) === JSON.stringify(trait.synonyms)) &&
```
ctucker3, line 140: Will this be able to account for different ordering of the tags and synonyms?

ctucker3 (changes requested):

timparsons:
Merged via #155 into releases/0.4

PR #155, opened 2021-12-03
[BI-1249] Tags - Saving after <Cancel>

davedrp:
see https://breedinginsight.atlassian.net/browse/BI-1249

I noticed that Synonyms was having the same problem and for the same reason, so I fixed that as well.


timparsons (approved):

ctucker3 (approved):
There is a bug with the equality check for tags and synonyms where the same list with different ordered elements will be considered unequal. But approving since that is minor and we want to get this out for testing.

PR #158, opened 2021-12-06
[BI-1055] Save germplasm

ctucker3:
This includes some fixes to the front end of the germplasm import that I noticed as I was working on 1055 for the backend. 

It is mainly fixes to better error reporting and displaying of the data.

### DIFF
--- a/src/views/import/ImportGermplasm.vue
+++ b/src/views/import/ImportGermplasm.vue
@@ -56,9 +56,9 @@
         </ConfirmImportMessageBox>
       </template>
 
-      <template v-slot:importPreviewTable="currentImport">
+      <template v-slot:importPreviewTable="previewData">
         <report-table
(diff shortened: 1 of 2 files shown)


HMS17 (approved):

nickpalladino (commented):
Screenshot from 2021-12-08 16-28-50

Did a little spot checking. Are we going to use the multiple error display like in the trait import and is that something for the future?

ctucker3:
> Did a little spot checking. Are we going to use the multiple error display like in the trait import and is that something for the future?

I believe that the multiple error display is being added in this card: https://breedinginsight.atlassian.net/jira/software/c/projects/BI/boards/1?modal=detail&selectedIssue=BI-1074

We'll have to update the errors I added in this one to work with the multiple error return format at some point.

nickpalladino (approved):

PR #160, opened 2021-12-09
[BI-1254] git trait tags and synonyms default values

davedrp:
for https://breedinginsight.atlassian.net/browse/BI-1254

The back-end should be release/0.4


ctucker3 (approved):
Code looks good. I also tested it with the scenarios in the JIRA card and confirmed that those scenarios are no longer an issue.

nickpalladino (approved):

PR #192, opened 2022-02-24
[BI-1337] ISSUE: System Users Table - Role and Programs  columns not sortable

dmeidlin:
# Description
[BI-1337](https://breedinginsight.atlassian.net/jira/software/c/projects/BI/boards/1?modal=detail&selectedIssue=BI-1337)

The roles and programs columns are made sortable using teh backend for the system users table.

# Dependencies
release/0.5 bi-api

# Testing
see the card

# Checklist:


src/breeding-insight/model/User.ts
```diff
@@ -37,9 +38,11 @@ export class User {
       this.roleName = role.name;
     }
     this.programRoles = programRoles;
+    this.programList = this.programRoles ? this.programRoles.map((x: ProgramUser = new ProgramUser() ) => x.program!.name).join(", ") : "";
```
ctucker3, line 41: In this `map` method, why is x set to a default of `new ProgramUser()`? Looking at the model, it looks like `program` will be null if you use that default, so your `x.program!.name` will throw an error in that case.

ctucker3 (changes requested):
Small clarification needed, I left an inline comment. 

Also, was there any work that was needed on the backend, or were all these fields already sortable on the backend?

davedrp (commented):
Happy Path Test passed

src/breeding-insight/model/User.ts
dmeidlin: Good point! fixed.
The fields were already sortable on the backend, so there isn't any change to bi-api.

ctucker3 (approved):

davedrp (approved):

PR #194, opened 2022-02-25
[BI-1356] Ordinal Nominal Switching Errors

HMS17:
# Description
**Story:** [ISSUE: New/Edit Term - Switching between scale classes, specifically ordinal and nominal, not working properly](https://breedinginsight.atlassian.net/browse/BI-1356)

Refactoring of Ordinal/Nominal display to get rid of anti-pattern, fix problem with ordinal/nominal category fields not displaying on first selection, and make code more robust.

- Functionality for modifying category "data" prop was moved from child component CategoryTraitForm.vue to parent component BaseTraitForm in order to avoid anti-pattern problems.
- Categories are now being passed in as categories (which is set to this.trait.scale.categories) due to the nested this.trait.scale.categories prop messing with reactivity
- Switching between ordinal/nominal now effectively hides the "value" fields rather than switching around item.value and item.label when changing scale class and maintaining a history. 
  - The category fields are consistent between ordinal and nominal display, with item.label being passed in as the category for both ValueRow.vue (used for nominal) and LabelValueRow.vue (used for ordinal). item.value is passed in as the value in LabelValueRow.vue, and is effectively "hidden" when switching from ordinal to nominal selection.
  - On save, if the scale class is nominal, the value and label are switched around in order to properly save to the database, and on load they are switched to maintain this more robust display.

# Dependencies
bi-api/release/0.5

# Testing

- Create new ontology term, select nominal, make sure one category displays and categories can be added and removed switch to ordinal and see that there are at minimum two fields and information is preserved
- Create new ontology term, select ordinal, make sure two categories display and categories can be added and removed switch to ordinal and see that there are at minimum two fields and information is preserved
- Save new ordinal term successfully
- Save new nominal term successfully
- Open existing nominal ontology term, check categories display properly, click edit, check display properly, save, check save properly
- Open existing ordinal ontology term, check categories display properly, click edit, check display properly, save, check save properly

# Checklist:

- [X] I have create/modified unit tests to cover this change
- [X] I have made corresponding changes to documentation
- [X] I have run TAF: [_\<link to TAF run>_](https://github.com/Breeding-Insight/taf/actions/runs/1959345494)

### DIFF
--- a/src/components/ontology/OntologyTable.vue
+++ b/src/components/ontology/OntologyTable.vue
@@ -494,7 +494,7 @@ export default class OntologyTable extends Vue {
   async saveTrait() {
     try {
       //For nominal traits switch back label value
-      if ((this.newTrait.scale) && (this.newTrait.scale.dataType) && (Scale.dataTypeEquals(this.newTrait.scale.dataType, DataType.Nominal)) && (this.newTrait.scale.categories)) {
+      if ((this.newTrait) && (this.newTrait.scale) && (this.newTrait.scale.dataType) && (Scale.dataTypeEquals(this.newTrait.scale.dataType, DataType.Nominal)) && (this.newTrait.scale.categories)) {
         this.newTrait.scale.categories.forEach(category => {
           category.value = category.label;
           category.label = undefined;
@@ -551,7 +551,7 @@ export default class OntologyTable extends Vue {
   async updateTrait(archiveStateChanged?: boolean) {
     try {
       //For nominal traits switch back label value
-      if ((this.editTrait.scale) && (this.editTrait.scale.dataType) && (Scale.dataTypeEquals(this.editTrait.scale.dataType, DataType.Nominal)) && (this.editTrait.scale.categories)) {
+      if ((this.editTrait) && (this.editTrait.scale) && (this.editTrait.scale.dataType) && (Scale.dataTypeEquals(this.editTrait.scale.dataType, DataType.Nominal)) && (this.editTrait.scale.categories)) {
         this.editTrait.scale.categories.forEach(category => {
           category.value = category.label;
           category.label = undefined;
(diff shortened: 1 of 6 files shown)


src/components/ontology/OntologyTable.vue
```diff
@@ -559,8 +569,17 @@ export default class OntologyTable extends Vue {
 
   async updateTrait(archiveStateChanged?: boolean) {
     try {
+      //For nominal traits switch back label value
+      let traitToSave = JSON.parse(JSON.stringify(this.editTrait));
+      if ((traitToSave) && (traitToSave.scale) && (traitToSave.scale.dataType) && (Scale.dataTypeEquals(traitToSave.scale.dataType, DataType.Nominal)) && (traitToSave.scale.categories)) {
```
timparsons, line 574: Minor thing, but it might be worth creating a method like `prepareScaleCategoriesForSave`, and have everything in this `if` block be in that method.  You could then call this new method from both `updateTrait` and `saveTrait`

timparsons (changes requested):
One small change request

src/components/ontology/OntologyTable.vue
HMS17: Good point, implemented!

timparsons (approved):

ctucker3 (changes requested):
Looks like this new code switches label and value for ordinal values. If I do, Test1 = 1 and Test2 = 2 on this code branch and qa-test they give two different results. Attaching screen shots of json from both. 

Local:
screencapture-jsonviewer-stack-hu-2022-03-14-14_29_51

QA Test:
screencapture-jsonviewer-stack-hu-2022-03-14-14_34_13

ctucker3 (approved):
Chatted with Heather. My reported error looks to have been a mistake with testing.

ctucker3 (approved):

PR #218, opened 2022-04-13
[BI-1375] - Make GIDs clickable links

HMS17:
# Description
**Story:** [BI-1375 - Make GIDs clickable links](https://breedinginsight.atlassian.net/browse/BI-1375)

Added GermplasmLink.vue component for germplasm GID links
Updated GermplasmDetails.vue to make parent GIDs clickable links (but not the current germplasm GID as it would be redundant)
Updated GermplasmTable.vue to make all GIDs clickable links
Changed germplasmUUID in GermplasmDetails.vue to a computed property in order to enable load of new germplasm information on clicking a parent GID

# Dependencies
[bi-api/feature/BI-1375](https://github.com/Breeding-Insight/bi-api/pull/177)

# Testing
- Open Germplasm Table
- Check that each GID, Female Parent GID, and Male Parent GID in a row is a link and leads to the correct germplasm details page
- Open Germplasm Details Page for a germplasm with male and female parent
- Parent GIDs should be displayed and each should be a clickable link to their respective germplasm details page
- Open Germplasm Details Page for a germplasm with female parent only
- Female parent GID should be displayed and should be a clickable link to its respective germplasm details page
- Open Germplasm Details Page for a germplasm with no parents
- No parent GID links should be displayed

# Checklist:

- [X] I have create/modified unit tests to cover this change
- [X] I have made corresponding changes to documentation
- [X] I have run TAF: [_\<link to TAF run>_](https://github.com/Breeding-Insight/taf/actions/runs/2173893697)


src/components/germplasm/GermplasmLink.vue
```diff
@@ -0,0 +1,46 @@
...
+  - distributed under the License is distributed on an "AS IS" BASIS,
+  - WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+  - See the License for the specific language governing permissions and
+  - limitations under the License.
+  -->
+
+<template>
+    <router-link v-if="this.germplasmUUID" v-bind:to="{name: 'germplasm-details', params: {programId: activeProgram.id, germplasmId: this.germplasmUUID}}">
+      {{ this.germplasmGID }}
+    </router-link>
```
ctucker3, line 21: I'm wondering for this if it would be better to have the GID not be a link if UUID isn't provided, rather than not showing the GID at all? 

I'm thinking that because from a user perspective it might be better to have a broken link rather than missing data in the case of a data error such as missing UUID.

src/views/germplasm/GermplasmDetails.vue
```diff
@@ -35,7 +35,18 @@
...
       <li><b>Source: </b> {{ germplasm.seedSource }}</li>
         <li><b>Pedigree: </b> {{ germplasm.additionalInfo.pedigreeByName }}</li>
-        <li><b>Pedigree GID(s): </b> {{ germplasm.pedigree }}</li>
+        <li><b>Pedigree GID(s): </b>
+          <GermplasmLink
+            v-if="germplasm.pedigree"
+            v-bind:germplasmUUID="Pedigree.parsePedigreeString(germplasm.additionalInfo.pedigreeByUUID).femaleParent"
+            v-bind:germplasmGID="Pedigree.parsePedigreeString(germplasm.pedigree).femaleParent"
+        > </GermplasmLink>
+          <template v-if="Pedigree.parsePedigreeString(germplasm.pedigree).maleParent">
```
ctucker3, line 44: In the GermplasmLink above you have the `v-if` directly in the component, but here you have it in a `template`. Is it possible to move this into the GermplasmLink for consistency?

src/views/germplasm/GermplasmTable.vue
```diff
@@ -26,18 +30,24 @@
...
+        > </GermplasmLink>
       </b-table-column>
       <b-table-column field="createdDate" label="Created Date" v-slot="props" :th-attrs="(column) => ({scope:'col'})">
         {{ props.row.data.additionalInfo.createdDate }}
       </b-table-column>
       <b-table-column field="createdBy.userName" label="Created By" v-slot="props" :th-attrs="(column) => ({scope:'col'})">
         {{ props.row.data.additionalInfo.createdBy.userName }}
       </b-table-column>
-      <b-table-column field="germplasmId" v-slot="props" :th-attrs="(column) => ({scope:'col'})">
+      <b-table-column field="externalReferences" v-slot="props" :th-attrs="(column) => ({scope:'col'})">
```
ctucker3, line 50: Wouldn't this field be something like `showDetails`?

ctucker3 (changes requested):

src/components/germplasm/GermplasmLink.vue
HMS17: Good point, changed to be text in the case of no UUID

src/views/germplasm/GermplasmTable.vue
HMS17: Based on discussion with Tim, confusion resolved by removing the field as it is unneeded here.

src/views/germplasm/GermplasmDetails.vue
HMS17: Right now I have the GermplasmLink as a one germplasm per link sort of thing rather than a more complicated component that includes multiple configurations of links. The v-if here is wrapped around the "/ <male parent GID>" part of the pedigree since the slash and gid only show up if there is a male parent. I could add some sort of boolean to pass in but I would still need a boolean on the outside for the slash lest I make the Germplasm Link more complicated by adding the slash. So what I have now seems to be the simplest route?

ctucker3: Ah gotcha, I missed the `/`. No need to change anything, thanks for the explanation!

ctucker3 (approved):

PR #234, opened 2022-05-27
[BI-1401] - Fixing showAll for pagination beyond page 2

HMS17:
# Description
**Story:** [“Show All” for programs pagination doesn’t work if user is on the last page of pagination
](https://breedinginsight.atlassian.net/browse/BI-1401)

Logic for toggling showAll failed on pages beyond page 2 (and only really worked on page 2 due to sheer luck of math). This change passes the correct page and pageSize information to the backend when showAll is toggled, regardless of starting page.

In addition, logic was added to ensure that when a new entry was added when the showAll toggle was selected, the pageSize increases by 1 so it does not spill over onto a second page.

This affects the following tables:

- AdminProgramsTable.vue
- AdminUsersTable.vue
- OntologyTable.vue
- ProgramLocationsTable.vue
- ProgramUsersTable.vue
- TraitsImportTable.vue

GermplasmTable.vue and GermplasmListTable.vue also converted to use BackendPaginationController instead of PaginationController for the sake of consistency and maintainability.

# Dependencies
bi-api/develop

# Testing
For aforementioned affected tables (at minimum one from the first list and then GermplasmTable.vue):

- Change page size such that there are greater than 2 pages
- Navigate to a page greater than 2
- Click showAll (toggling on)
- All values should be displayed and pagination should be on page 1
- Click showAll (toggling off)
- Previous page size should be restored with current page being page 1
- Click showAll (toggling on)
- Change page size
- List should display chosen page size on page 1 with showAll toggled off

# Checklist:

- [X] I have run TAF: [_\<link to TAF run>_](https://github.com/Breeding-Insight/taf/actions/runs/2496577400)

### DIFF
--- a/src/breeding-insight/model/view_models/BackendPaginationController.ts
+++ b/src/breeding-insight/model/view_models/BackendPaginationController.ts
@@ -63,9 +63,19 @@ export class BackendPaginationController {
     this.showAll = !this.showAll;
   }
 
-  static getPaginationSelections(currentPage: number, pageSize: number): PaginationQuery {
+  static getPaginationSelections(currentPage: number, pageSize: number, showAll: boolean): PaginationQuery {
+    /*
+    if (showAll==undefined) showAll = false; //todo possibly temporary if
     return new PaginationQuery(
-      currentPage, pageSize, false);
+      currentPage, pageSize, showAll); //HERE TODO
+     */
+    if (showAll) {
+      return new PaginationQuery(
+          1, pageSize, showAll); //0 or 1
+    } else {
+      return new PaginationQuery(
+          currentPage, pageSize, showAll);
+    }
   }
 
 }
--- a/src/breeding-insight/model/view_models/BackendPaginationController.ts
+++ b/src/breeding-insight/model/view_models/BackendPaginationController.ts
@@ -67,7 +67,7 @@ export class BackendPaginationController {
     if (this.showAll) {
       this.updatePageSizeVals(this.lastPageSize);
     } else {
-      this.updatePageSizeVals(totalCount); //todo,need to get count updating when show all active
+      this.updatePageSizeVals(totalCount);
       this.updatePage(1);
     }
     this.showAll = !this.showAll;
--- a/src/breeding-insight/model/view_models/BackendPaginationController.ts
+++ b/src/breeding-insight/model/view_models/BackendPaginationController.ts
@@ -78,4 +78,12 @@ export class BackendPaginationController {
           currentPage, pageSize, showAll);
   }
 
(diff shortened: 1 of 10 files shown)


dmeidlin (commented):
When I go through the steps listed in the testing criteria the pagination controls behave as expected except for the case when the total number of records in the table equals one of the possible page sizes. 

If the total number is 51 then toggling on "show all" will list all on a single page with the per page drop down menu blank and the show-all button will have the tertiary info background color. Selecting a page size from the drop down will update the table displaying the page size in the drop down and the show-all button will have  a gray background.

If the total number is 50 than the controls still update the table as in the prior case but when "show all" is toggled on the the page size displayed is shown as 50 (whereas the drop down would normally be blank) and selecting 50 as the page size does not toggle off "show all" (when otherwise this action would).  

The page size is correct, so functionally it's fine, and this is admittedly a minor point. It would be nice, though, if the controls would behave the same regardless of the table size.

src/breeding-insight/model/view_models/BackendPaginationController.ts
```diff
@@ -56,16 +65,25 @@ export class BackendPaginationController {
 
   toggleShowAll(totalCount: number) {
     if (this.showAll) {
-      this.updatePageSize(this.lastPageSize);
+      this.updatePageSizeVals(this.lastPageSize);
     } else {
-      this.updatePageSize(totalCount);
+      this.updatePageSizeVals(totalCount); //todo,need to get count updating when show all active
```
nickpalladino, line 70: just wondering about the comment here.

HMS17: I think that was left over from when I was trying to get the showAll to work on adding a new value (since the totalCount wouldn’t update and you’d get a second page despite showAll being toggled). Removed!

HMS17:
> When I go through the steps listed in the testing criteria the pagination controls behave as expected except for the case when the total number of records in the table equals one of the possible page sizes.
> 
> If the total number is 51 then toggling on "show all" will list all on a single page with the per page drop down menu blank and the show-all button will have the tertiary info background color. Selecting a page size from the drop down will update the table displaying the page size in the drop down and the show-all button will have a gray background.
> 
> If the total number is 50 than the controls still update the table as in the prior case but when "show all" is toggled on the the page size displayed is shown as 50 (whereas the drop down would normally be blank) and selecting 50 as the page size does not toggle off "show all" (when otherwise this action would).
> 
> The page size is correct, so functionally it's fine, and this is admittedly a minor point. It would be nice, though, if the controls would behave the same regardless of the table size.

The bug exists in qa-test as well, but added a fix.

nickpalladino (approved):

PR #285, opened 2022-12-05
[BI-1615] - Add termType to Trait DB Table

HMS17:
# Description
**Story:** [BI-1615 - Add termType to Trait DB Table](https://breedinginsight.atlassian.net/browse/BI-1615)

NOTE: This is forked from BI-1614, so changelist will include changes from there until BI-1614 is merged into develop

Small improvements off BI-1614 to ensure term types actually display
Created EnumUtils to handle methods for converting between enum key and value for ease of converting between the user-friendly display value and the backend storage value, and utilized them for term types

NOTE: Sorting could get weird if the alphabetical order differs between the user friendly display (ie Germplasm Passport) and backend storage value (GERM_PASSPORT). Not a problem here, but something to keep in mind. Also the difference between display and backend and the fact that there are a few distinct types means that if filtering was to be implemented for term types in the future it should probably be via drop down rather than text box.

# Dependencies
[bi-api/BI-1615](https://github.com/Breeding-Insight/bi-api/pull/231)

# Testing

- Open ontology table
- Check that Term Type column exists and is populated with values
- Check that Term Type column can be sorted
- Click new term, check that there is now a Term Type dropdown with "Phenotype" as default and a dropdown that lists "Phenotype", "Germplasm Attribute", and "Germplasm Passport"
- Select term type and save, check that new ontology term displayed in table with correct term type 
- Click 'Show Details' for one germplasm, check side panel display includes Term Type and Term Type is correct value
- Click edit term, check that Term Type dropdown value matches the Term Type in table
- Change Term Type and check that new value is saved and displayed correctly in table
- Check that Term Type column also exists in Archived Ontology table and not just Active

# Checklist:

- [ ] I have run TAF: [_\<link to TAF run>_](https://github.com/Breeding-Insight/taf/actions/runs/3624530763)

### DIFF
--- a/src/breeding-insight/model/Sort.ts
+++ b/src/breeding-insight/model/Sort.ts
@@ -55,7 +55,8 @@ export enum OntologySortField {
   MethodDescription = 'methodDescription',
   ScaleClass = 'scaleClass',
   ScaleName = 'scaleName',
-  entityAttributeSortLabel = 'entityAttribute'
+  entityAttributeSortLabel = 'entityAttribute',
+  TermType = 'termType'
 }
 
 export class OntologySort {
--- a/src/breeding-insight/model/Trait.ts
+++ b/src/breeding-insight/model/Trait.ts
@@ -88,6 +88,7 @@ export class Trait {
     }
     this.fullName = fullName;
     this.isDup = isDup;
+    this.termType = termType;
   }
 
   static assign(trait: Trait): Trait {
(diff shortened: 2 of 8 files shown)


dmeidlin (changes requested):
All features work as expected. The "Term Type" field in both the edit and new ont. term forms has an asterisk (denoting required field) and is listed at the top above the Name and Description fields. While the frontend and backend checks should stay in place, the asterisk is superfluous and should be removed since the dropdown prevents inputing unsupported or null values. Also, the Term Type field should be below the Description field and above the synonym field to be more consistent with the order of the fields in the details summary and table column headings.

PR #354, opened 2023-12-01
[BI-1867] - Ontology: Increase name length to 16 characters

HMS17:
# Description
**Story:**  [ BI-1867 - Ontology: Increase name length to 16 characters ](https://breedinginsight.atlassian.net/browse/BI-1867)

Modified front end logic to increase ontology name length limit from 12 to 16 characters.

# Dependencies
[bi-api: feature/BI-1867](https://github.com/Breeding-Insight/bi-api/pull/317)

# Testing
In both ontology import and manual entry of ontology:
- Try to add an ontology term with 16 characters, should succeed and new entry show up in ontology table 
- Try to add an ontology term with 17 characters, confirm that an error message indicating that character limit must be 16 or less shows up and import is blocked and no new terms are added to ontology table

# Checklist:

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "bi-web",
-  "version": "v0.9.0+633",
+  "version": "v0.9.0+635",
   "private": true,
   "scripts": {
     "build": "node $npm_package_config_task_path/build.js --dev-audit-level=critical --prod-audit-level=none",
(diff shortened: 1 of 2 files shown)


mlm483 (approved):
Works as expected.

I noticed that the error message when the name is too long says "Name must be less than 16 characters", which isn't quite right. Might be worth changing [L103 in BaseFieldWrapper.vue](https://github.com/Breeding-Insight/bi-web/blob/10fb0334dcd78fec2ac8c62a37ff5384774596e4/src/components/forms/BaseFieldWrapper.vue#L103) to say "...must be no longer than..." instead. That could also be a separate change.
Screenshot 2023-12-04 at 12 48 04 PM

davedrp (approved):
passed developer testing

PR #380, opened 2024-05-16
BI-2105 (Accessibility: Header and Navigation)

davedrp:
# Description
[BI-2105](https://breedinginsight.atlassian.net/browse/BI-2105) - Accessibility: Header and Navigation

_changes made:
EXPECTED BHAVIOUR and how it was implemented_

**The program name and System Administration headers on the left sidebar should NOT be transformed to all caps by the text-transform property.**
-  in `main.scss` added the class selector `.menu-label.customize` to overwrite the buefy `.menu-label` class selector.

**The user status menu dropdown button should have an associated text label in the HTML.** 
- added the attribute `aria-label="User Status"`to the user status button

 **The user status menu dropdown div should either be a compliant role for the aria-haspopup state OR the aria-pressed state should be used for the enclosed button role.**
- IMPORTANT NOTE:  Could not correct this problem because it is within the Buefy code for a dropdown.
 
**The program dropdown menu form should have an associated label.** 
- added an HTML label with `class="is-sr-only"` so it is only visible to Screen Readers.

# Dependencies
_Please include any dependencies to other code branches, testing configurations, scripts to be run, etc._

# Testing
Confirm that the accessibility issues have been resolved by using the Chrome plug-in - https://chromewebstore.google.com/detail/siteimprove-accessibility/djcglbmbegflehmbfleechkjhmedcopn?pli=1

# Checklist:

[BI-2105]: https://breedinginsight.atlassian.net/browse/BI-2105?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/assets/scss/main.scss
+++ b/src/assets/scss/main.scss
@@ -90,7 +90,9 @@ $medium-modal-body-padding: 0 3em 2em 3em;
 @import "~bulma/sass/utilities/_all";
 
 // Customize buefy if you want to here
-.menu-label.customize {
+
+//see https://alfa.siteimprove.com/rules/sia-r72
+.menu-label.customize_for_accessibility {
   text-transform: none;
   font-size: 0.8em;
   letter-spacing: 2px;
--- a/src/assets/scss/main.scss
+++ b/src/assets/scss/main.scss
@@ -90,7 +90,11 @@ $medium-modal-body-padding: 0 3em 2em 3em;
 @import "~bulma/sass/utilities/_all";
 
 // Customize buefy if you want to here
-
+.menu-label.customize {
+  text-transform: none;
+  font-size: 0.8em;
+  letter-spacing: 2px;
+}
 
 // Update some of Bulma's component variables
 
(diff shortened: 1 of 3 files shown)


src/assets/scss/main.scss
```diff
@@ -90,7 +90,11 @@ $medium-modal-body-padding: 0 3em 2em 3em;
 @import "~bulma/sass/utilities/_all";
 
 // Customize buefy if you want to here
-
+.menu-label.customize {
```
nickpalladino, line 93: Can this be named differently like something related to accessibility.

mlm483 (approved):

src/assets/scss/main.scss
davedrp: Done.

src/assets/scss/main.scss
```diff
@@ -91,6 +91,12 @@ $medium-modal-body-padding: 0 3em 2em 3em;
 
 // Customize buefy if you want to here
 
+//see https://alfa.siteimprove.com/rules/sia-r72
+.menu-label.customize_for_accessibility {
```
nickpalladino, line 95: Sorry to be a pain but could you use - for separators in the name so it's consistent with the convention in the rest of the file.

nickpalladino (changes requested):

src/assets/scss/main.scss
davedrp: Done. (Good catch)

PR #404, opened 2024-09-25
BI-1933 - Erroneous message: No germplasm are currently defined for this program 

mlm483:
# Description
**Story:** [BI-1933](https://breedinginsight.atlassian.net/browse/BI-1933)

A common pattern in bi-web is to have a get{Entities} method for the data in a table view that is called within the `mounted()` method of the component, while also having a `paginationChanged` method with a `@Watch` annotation reference an `PaginationController` that can also call get{Entities}. In some (but not all) cases, the `PaginationController` instance is mutated in the `mounted()` method. This resulted in the get{Entities} method being called twice on initial page load, and (maybe) because of the way Vue handles reactivity, the "empty" table message would sometimes flash on screen. The larger issue though was that the duplicate calls were unnecessary, and since `paginationChanged` already has short-circuiting logic built in to ensure get{Entities} is called only when the request parameters are changed, a good solution was to remove the call to get{Entities} in `mounted()` and directly call `paginationChanged` instead.

Additionally, some components were calling `updatePagination()` from `mounted()`, which is also called *when needed* by `paginationChanged()`, so I removed those calls from `mounted()`.

# Testing
1. Import a large germplasm file (1000+ records) and ensure the "No germplasm are currently defined for this program." message does not flash on screen while the germplasm are loading into the table view. Note that this is hard to reproduce even when present, I recommend focusing on steps 2. and 3. instead.
2. Ensure all tables load data.
3. In the dev tools network tab, ensure duplicate (identical) calls are not being made on initial page load.
These are the affected table components and URL paths where they can be tested:

| File                             | Route                                                     |
|----------------------------------|-----------------------------------------------------------|
| AdminProgramsTable.vue           | /admin/programs                                           |
| AdminUsersTable.vue              | /admin/users                                              |
| ExperimentsObservationsTable.vue | /programs/{programId}/experiments-observations            |
| GermplasmListsTable.vue          | /programs/{programId}/germplasm/germplasm-lists           |
| GermplasmTable.vue               | /programs/{programId}/germplasm/germplasm-all             |
| OntologyTable.vue                | /programs/{programId}/ontology/[active \| archived]-terms |
| ProgramLocationsTable.vue        | /programs/{programId}/program-administration/locations    |
| ProgramUsersTable.vue            | /programs/{programId}/program-administration/users        |
| TraitsImportTable.vue*           | /programs/{programId}/import/ontology                     |

*To test the TraitsImportTable component, import ontology and make sure the preview displays correctly.

# Checklist:

[BI-1933]: https://breedinginsight.atlassian.net/browse/BI-1933

### DIFF
--- a/src/components/admin/AdminProgramsTable.vue
+++ b/src/components/admin/AdminProgramsTable.vue
@@ -311,7 +311,6 @@ export default class AdminProgramsTable extends Vue {
   }
 
   mounted() {
-    this.updatePagination();
     this.getSpecies();
     this.paginationChanged();
   }
@@ -333,7 +332,7 @@ export default class AdminProgramsTable extends Vue {
 
   @Watch('paginationController', { deep: true})
   paginationChanged() {
-    let currentCall = this.paginationController.currentCall
+    let currentCall = this.paginationController.currentCall;
     let paginationQuery = this.paginationController.getPaginationSelections();
     if(currentCall && currentCall!.page == paginationQuery.page && currentCall!.pageSize == paginationQuery.pageSize && currentCall!.showAll == paginationQuery.showAll) {
       return;
--- a/src/components/admin/AdminProgramsTable.vue
+++ b/src/components/admin/AdminProgramsTable.vue
@@ -312,8 +312,8 @@ export default class AdminProgramsTable extends Vue {
 
   mounted() {
     this.updatePagination();
-    this.getPrograms();
     this.getSpecies();
+    this.paginationChanged();
   }
 
   setSort(field: string, order: string) {
@@ -348,6 +348,7 @@ export default class AdminProgramsTable extends Vue {
   }
 
   getPrograms() {
+    this.programsLoading = true;
     ProgramService.getAll(this.paginationController.currentCall, this.programSort).then(([programs, metadata]) => {
 
       // Check that our most recent query is this one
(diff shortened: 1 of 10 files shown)


src/components/admin/AdminProgramsTable.vue
```diff
@@ -311,9 +311,8 @@ export default class AdminProgramsTable extends Vue {
   }
 
   mounted() {
-    this.updatePagination();
-    this.getPrograms();
     this.getSpecies();
```
nickpalladino, line 314: Does having this getSpecies call before the programsLoading flag being set in getPrograms impact the timing at all?

mlm483: I don't observe any issues, two things to note:
- `programsLoading` is initialized to `true` so it doesn't matter if there's a delay before `getPrograms` is called,
- the `ExpandableTable` will show as loading until both species and programs have loaded (see snippet below). 

```
 <ExpandableTable
      v-bind:records.sync="programs"
      v-bind:loading="this.speciesLoading || this.programsLoading"
```

nickpalladino (approved):

davedrp (approved):

mlm483:
I'll need to create a new feature branch to merge these changes into release/1.0, this branch was based off of develop after some v1.1 stories were already merged.

PR #447, opened 2025-07-24
[BI-2630] - Improve Create Sub-Entity Pop-Up

HMS17:
# Description
**Story:** [BI-2630 - Improve Create Sub-Entity Pop-Up](https://breedinginsight.atlassian.net/browse/BI-2630)

Changes to improve and add validations for modal for creating new sub-entities.

- Changed validation on repeated measures range to permit values 2-50 and updated corresponding text on modal
- Added logic to retrieve entity names stored in program and entity/dataset names currently in experiment
- Updated sub-entity modal field for dataset name to suggest entity names stored in the program that are not currently sub-entity names in the experiment
- Added validation for sub-entity creation to prevent creating a sub-entity with the same name as an existing sub-entity in the experiment
- Changed font color in sub-entity creation modal to meet WCAG AA color contrast standards

(Note: The suggested entity names are the program entity names as stored in the /observation-levels endpoint when creating an entity as part of an ontology term, not all the experiment-associated entity/sub-entity names. It might be worth considering whether experiment-associated entity/sub-entity names should be stored there rather than just in additionalInfo)

# Dependencies
bi-api: develop

# Testing
In experiment dataset view
- Click Manage Experiment > Create Sub-entity Dataset
- Check that the text beneath the repeated measures field says "Maximum expected between 2-50"
- Check that a user can only create a sub-entity with repeated measures within the range of 2-50 and a validation error occurs otherwise
- Click on sub-entity name field and check that a dropdown of entity names appears that is drawn from program entity names minus any entity/dataset names in the current experiment (try making a entity name when creating an ontology term that is the same as a dataset name for the experiment to test this)
- Check that if one enters a sub-entity name that is already the name of an entity in the current experiment (case insensitive), on clicking save an error message banner is displayed and a new sub-entity is not added

# Checklist:

- [ ] I have run TAF: _\<link to TAF run>_ (taf presently too broken)

### DIFF
--- a/src/components/modals/SubEntityDatasetModal.vue
+++ b/src/components/modals/SubEntityDatasetModal.vue
@@ -69,7 +69,7 @@
                 v-model="newSubEntity.repeatedMeasures"
                 v-bind:validations="validations.repeatedMeasures"
                 v-bind:field-name="'Repeated Measures'"
-                v-bind:field-help="'Maximum expected'"
+                v-bind:field-help="'Maximum expected between 2-50'"
               />
             </div>
           </div>
@@ -116,6 +116,8 @@ export default class SubEntityDatasetModal extends Vue {
   experiment!: Trial;
   @Prop()
   defaultObservationLevel?: string;
+  @Prop()
+  existingDatasetNames!: string[];
 
   // Reactive, private (would not be reactive if declared without initial values).
   private newSubEntity: SubEntityDatasetNewRequest = new SubEntityDatasetNewRequest();
@@ -130,7 +132,7 @@ export default class SubEntityDatasetModal extends Vue {
     repeatedMeasures: {
       required,
       integer,
-      between: between(1, 50),  // Note: capped at 50 for performance considerations.
+      between: between(2, 50),  // Note: capped at 50 for performance considerations
     }
   }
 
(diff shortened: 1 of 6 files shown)


davedrp (approved):
Test passed.
Code looks good.

Great work (as usual)!

src/components/modals/SubEntityDatasetModal.vue
```diff
@@ -69,7 +69,7 @@
                 v-model="newSubEntity.repeatedMeasures"
                 v-bind:validations="validations.repeatedMeasures"
                 v-bind:field-name="'Repeated Measures'"
-                v-bind:field-help="'Maximum expected'"
+                v-bind:field-help="'Maximum expected between 2-50'"
```
nickpalladino: I think it would be valid to have 1 repeated measure, the backend allows it. It would result in one sub-observation unit being created per exp unit. Is this something you talked with Shawn about?

src/components/modals/SubEntityDatasetModal.vue
```diff
@@ -130,7 +132,7 @@ export default class SubEntityDatasetModal extends Vue {
     repeatedMeasures: {
       required,
       integer,
-      between: between(1, 50),  // Note: capped at 50 for performance considerations.
+      between: between(2, 50),  // Note: capped at 50 for performance considerations
```
nickpalladino: See previous comment for minimum bound change

src/components/modals/SubEntityDatasetModal.vue
```diff
@@ -142,6 +144,14 @@ export default class SubEntityDatasetModal extends Vue {
   }
 
   async invokeCreate(){
+    //Check if sub-entity name is already in experiment
+    let nameAlreadyInExp = this.existingDatasetNames.map(y => y.toLowerCase()).includes(this.newSubEntity.name.toLowerCase());
```
nickpalladino, line 148: We might want to have this check done in the backend in the `POST /programs/{programId}/experiments/{experimentId}/dataset` endpoint. I'd be ok having that in another card.

nickpalladino (changes requested):

src/components/modals/SubEntityDatasetModal.vue
HMS17: Changed to 1

src/components/modals/SubEntityDatasetModal.vue
HMS17: Changed to 1

nickpalladino (approved):

PR #459, opened 2026-03-06
[BI-2778] Deprecate additionalinfo usage for level and use obs unit lvl

HMS17:
# Description
**Story:** [BI-2778 - Deprecate additionalinfo usage for level and use obs unit level](https://breedinginsight.atlassian.net/browse/BI-2778)

Changes to ensure front end display of obs lvl is capitalized.

# Dependencies
bi-api: [feature/BI-2778](https://github.com/Breeding-Insight/bi-api/pull/500)

# Testing
see bi-api

# Checklist:

- [X] I have either updated the source of truth or arranged for update with product owner if needed https://breedinginsight.atlassian.net/wiki/spaces/BI/pages/1559953409/Source+of+Truth

### DIFF
--- a/src/components/modals/SubEntityDatasetModal.vue
+++ b/src/components/modals/SubEntityDatasetModal.vue
@@ -46,7 +46,7 @@
         <template v-slot="validations">
           <div class="message is-success">
             <div class="message-body has-text-dark">
-              Prepare a dataset for repeated observations within {{ defaultObservationLevel }}.
+              Prepare a dataset for repeated observations within {{ StringFormatters.toStartCase(defaultObservationLevel) }}.
             </div>
           </div>
           <div class="columns is-multiline mb-4">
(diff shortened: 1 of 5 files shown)


src/components/modals/SubEntityDatasetModal.vue
```diff
@@ -93,8 +93,14 @@ import DataForm from "@/components/forms/DataForm.vue";
...
 import {SubEntityDatasetNewRequest} from "@/breeding-insight/model/SubEntityDatasetNewRequest";
 import {DataFormEventBusHandler} from "@/components/forms/DataFormEventBusHandler";
+import {StringFormatters} from "../../breeding-insight/utils/StringFormatters";
 
 @Component({
+  computed: {
+    StringFormatters() {
+      return StringFormatters
+    }
+  },
```
jloux-brapi, line 103: Now that I'm seeing this for a second time, I'm wondering if there's a more elegant way to do this without this same boilerplate code.

If we can create some sort of global obj/utility that can be used by any component I think we could do that, but I do know it's more work.

For now, I think what you have is fine, but we should start to think about this if the use of StringFormatters continues to get larger.

jloux-brapi (approved):

PR #466, opened 2026-05-06
[BI-2806] Route Experiments using trialDbId, use trialDbId to get metadata

jloux-brapi:
# Description
**Story:** [BI-2806](https://breedinginsight.atlassian.net/browse/BI-2806)

These changes support the removal of BrAPITrials from the ProgramCache.

The Experiment table will now route experiments using their BrAPITrialDbId, and the metadata grab associated will also use the BrAPITrialDbId.

# Dependencies
[bi-api PR](https://github.com/Breeding-Insight/bi-api/pull/516)

# Testing
Regression on just about anything Experiment related, including:

- Experiment creation
- Experiment deletion
- Experiment updates
- Experiment Collaborator checks
- Experiment viewing
- Dataset viewing
- Dataset creation
- Dataset appending

Verify there are no log messages related to caching trials for a program by looking at logs once above tests have been completed.

# Checklist:

- [X] I have either updated the source of truth or arranged for update with product owner if needed https://breedinginsight.atlassian.net/wiki/spaces/BI/pages/1559953409/Source+of+Truth

[BI-2806]: https://breedinginsight.atlassian.net/browse/BI-2806?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/breeding-insight/service/ExperimentService.ts
+++ b/src/breeding-insight/service/ExperimentService.ts
@@ -68,16 +68,13 @@ export class ExperimentService {
         if (!trial) {
             return ResultGenerator.err(new Error('Missing or invalid trial'));
         }
-        if (!trial.externalReferences) {
-            return ResultGenerator.err(new Error('Trial is missing external references'));
+
+        let trialDbId = trial.trialDbId;
+        // Throw if trial is missing trialDbId.
+        if (!trialDbId) {
+            return ResultGenerator.err(new Error("Trial is missing a brapi dbId."));
         }
-        // Try to get the /trials external reference.
-        let externalReferenceId = BrAPIUtils.getBreedingInsightId(trial.externalReferences, '/trials');
-        // Throw if trial is missing ExternalReferenceId.
-        if (!externalReferenceId) {
-            return ResultGenerator.err(new Error("Trial is missing external reference."));
-        }
-        return await ExperimentDAO.getDatasetMetadata(programId, externalReferenceId);
+        return await ExperimentDAO.getDatasetMetadata(programId, trialDbId);
     }
     
     static async getUnassignedCollaboratorsByExperiment(programId: string | undefined, experimentId: string): Promise<Result<Error, Collaborator[]>> {
--- a/src/components/experiments/ExperimentsObservationsTable.vue
+++ b/src/components/experiments/ExperimentsObservationsTable.vue
@@ -43,7 +43,7 @@
       v-on:search="initSearch"
     >
       <b-table-column label="Title" field="name" cell-class="fixed-width-wrapped" sortable v-slot="props" :th-attrs="(column) => ({scope:'col'})" searchable>
-        <router-link v-bind:to="{name: 'experiment_dataset', params: {datasetId: getDefaultDataset(props.row.data).id, programId: activeProgram.id, experimentId: BrAPIUtils.getBreedingInsightId(props.row.data.externalReferences,'/trials')}}">
+        <router-link v-bind:to="{name: 'experiment_dataset', params: {datasetId: getDefaultDataset(props.row.data).id, programId: activeProgram.id, experimentId: props.row.data.trialDbId}}">
           {{ props.row.data.trialName }}
         </router-link>
 


davedrp (approved):

HMS17 (approved):

humsika (approved):
Everything looks good to me.

src/components/experiments/ExperimentsObservationsTable.vue
nickpalladino, line 25: The endpoint `GET /experiments/{experimentId}/export` was updated to expect trialDbId for experimentId but the table modal still passes in the DeltaBreed UUID so exporting an experiment from the Experiments list download links will result in a not found error.

src/breeding-insight/service/ExperimentService.ts
nickpalladino, line 29: Wonder if we should rename experimentId to trialId in these service and DAO methods to make it more clear which id is expected?

src/components/experiments/ExperimentsObservationsTable.vue
```diff
@@ -43,7 +43,7 @@
       v-on:search="initSearch"
     >
       <b-table-column label="Title" field="name" cell-class="fixed-width-wrapped" sortable v-slot="props" :th-attrs="(column) => ({scope:'col'})" searchable>
-        <router-link v-bind:to="{name: 'experiment_dataset', params: {datasetId: getDefaultDataset(props.row.data).id, programId: activeProgram.id, experimentId: BrAPIUtils.getBreedingInsightId(props.row.data.externalReferences,'/trials')}}">
+        <router-link v-bind:to="{name: 'experiment_dataset', params: {datasetId: getDefaultDataset(props.row.data).id, programId: activeProgram.id, experimentId: props.row.data.trialDbId}}">
```
nickpalladino, line 46: This will also change the id that is passed to `GET /experiments/{experimentId}/collaborators` to trialId but existing records in the experiment_program_user_role table will have the deltabreed experiment UUID rather than the trialId stored in experiment_id. It looks like the only records in prod are in a test program so it may not be an issue as any new collaborators would be stored with the trialId. If we did have to do anything, I think it should go in a new card.

nickpalladino (changes requested):

src/components/experiments/ExperimentsObservationsTable.vue
jloux-brapi: Good catch, I think I forgot to test that feature.  I will fix this.

src/breeding-insight/service/ExperimentService.ts
jloux-brapi: We could, altho functionally I kind of think they largely mean the same thing as of these changes.

src/components/experiments/ExperimentsObservationsTable.vue
nickpalladino: sounds good

src/breeding-insight/service/ExperimentService.ts
nickpalladino: ya I guess I'm thinking to avoid issues of passing in one type of id and thinking it's another, although renaming also wouldn't really fix that in a robust way. Could also type it to a specific id type rather than using a string to be more self documenting.

src/components/experiments/ExperimentsObservationsTable.vue
jloux-brapi: I decided in the interest of moving this stuff forward into the epic branch, I will address this in the following [bug fix](https://breedinginsight.atlassian.net/browse/BI-2934)

src/components/experiments/ExperimentsObservationsTable.vue
jloux-brapi: All good points.  I made [the following ticket](https://breedinginsight.atlassian.net/browse/BI-2935) and tagged it in the epic for review towards the end of the cache removal implementation.  I think it's possible it doesn't need to be implemented but sounds important enough to be captured somewhere.

src/breeding-insight/service/ExperimentService.ts
jloux-brapi: My opinion is that an BI generated ID that's being stored in exrefs and then used should be called something like `biEntityExRef`.  This I think would avoid confusion.

But yea IMO our end goal should be to eliminate this kind of confusion by having dedicated entities on the brapi side that meet those needs.  I know it's not all there right now, but that's where I think it should be.

nickpalladino (approved):

